PR #97, opened 2015-01-07
Automatic generation of the scenario test function names.

### LINKED ISSUE #89: DRY for function names when applying decorators
Tests are created using ```@scenario``` decorator and after that the function has to be written (with unique function name). So basically test name exist in 3 places, feature file, decorator and (kind of) in function name. Instead something like:
```
@scenario(....)
def _():
    pass
```
could be used to generate functions automatically searching for unique names.

bubenkoff:
closes #89

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,6 +93,31 @@ test_publish_article.py:
         assert article.is_published
 
 
+Less code for scenario test function
+------------------------------------
+
+pytest-bdd `scenario` decorator can manage the scenario function name for you automatically to reduce the amount of
+code you need to write:
+
+.. code-block:: python
+
+    from pytest_bdd import scenario
+
+    @scenario('publish_article.feature', 'Publishing the article')
+    def _():
+        pass
+
+As a result, module will get the function:
+
+
+.. code-block:: python
+
+
+    def test_publishing_the_article():
+        """publish_article.feature: Publishing the article"""
+
+so docstring is also automatically generated.
+
 Step aliases
 ------------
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -93,6 +93,31 @@ test_publish_article.py:
...
+    def _():
+        pass
+
+As a result, module will get the function:
+
+
+.. code-block:: python
+
+
+    def test_publishing_the_article():
```
olegpidsadnyi: So you won't see the "test_publishing_the_article" explicitly? How do you copy-paste it to run py.test -k test_publishing_the_article then?

pytest_bdd/steps.py
```diff
@@ -46,6 +47,8 @@ def article(author):
 )
 from .parsers import get_parser
 
+PYTHON_NAME_RE = re.compile('[^0-9a-zA-Z\s]+')
```
olegpidsadnyi: \d\w\S?

tests/args/regex/test_args.py
```diff
@@ -20,17 +20,17 @@
 
 
 @scenario_args('Every step takes a parameter with the same name')
-def test_steps():
+def _():
```
olegpidsadnyi: in case of given you cen use it as a fixture later, right?

@given('I have a foo')
def foo():
    return 1

@then('Foo should be 1' )
def _(foo):
     assert foo == 1

README.rst
```diff
@@ -126,6 +151,33 @@ default author.
         And there is an article
 
 
+Less code for step functions
+----------------------------
+
+pytest-bdd `when` and `then` decorators can manage the function names for you automatically (as well as `scenario`),
+and it's not necessary to choose unique function names manually, instead you can have for example `_` as a function
```
olegpidsadnyi: Do you expect any problems with habits of importing ugettext as _?

PR #102, opened 2015-03-15
faster travis builds

santagada:
As explained in: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

By setting sudo to false it will use the new docker infrastructure.


bubenkoff:
great!

Issue #92, opened 2014-12-30
Not implemented scenarios are silently omitted.


spinus:
When the test is not implemented they are not executed without a warning.
I can imagine scenario like:
1. I have a feature file with scenario "This is my scenario"
2. I have implemented scenario with:
```
@scenario("my.feature", "This is my scenario")
```
3. I forgot to add dot to a feature file and I'm adding now.
4. If I forgot to fix implementation the scenario is lost silently.

It would be nice to have a warning/error (pytest command line option?) to say that some tests are not implemented (or maybe even that some steps are never used?).

bubenkoff:
i think this is a duplication of #90, please check

spinus:
@bubenkoff seems like it can be duplication if somebody could use only ```scenarios```. But if in one file you use ```@scenario``` and not ```scenarios``` it's still the case I think.

bubenkoff:
then i don't understand...
if you have this:

```python
@scenario('some.feature', 'Some Scenario')
def test_something():
    pass
```
it will fail immediately if you forgot something
May be you mean that it will not fail if the scenario in the feature file is empty by itself?

spinus:
I'm not sure I'm talking about that in a good way. Let me use example.
``my_feature.feature``:
```
Feature: my feature
    Scenario: scenario1
        Given a
        When b
        Then c
```
and implementation:
``test_my_feature.py``:
```
# empty
```
When I run:
```
py.test
```
```
platform linux2 -- Python 2.7.6 -- py-1.4.26 -- pytest-2.6.4
plugins: bdd
collected 0 items
```
No error about not implemented test.

bubenkoff:
right, so you're about not bound scenarios problem
but in the issue description it was:

2. I have implemented scenario with:

@scenario("my.feature", "This is my scenario")

so i was confused.
The only realistic solution for me is the helper which we already have: `pytest-bdd generate-missing` command.
Also the `scenarios` mentioned in #90 will help a bit.
But it can not be solved completely: the entry point is the test, not the feature file, that's where it's `wrong`.
And it completely depends on which tests you've collected ATM, so you don't know if you've collected all tests or only subset of them (because pytest allows you to run only a certain folder or even just single test).
We could throw the warning, but it will look strange if you run single test while having big feature file

spinus:
Yes, sorry for this. I was writting crap. The title of the issue is good, but first explanation pretty bad, I think I've lost focus when I was writting it :).

[...]

Issue #90, opened 2014-12-30
@scenario decorator is an overkill


spinus:
Repeating full scenario phrase in feature file and then in ```@scenario``` decorator is a little cumbersome. As @bubenkoff said, that approach has few values over the implicit scenario detection. 
Some of them:
* assertion in scenario function (executed after scenario is run)
* scenarios can be parametrized and use fixtures
So removing @scenario decorator won't work as we want to have those features.
But maybe for simple feature files without a lot of requirements we can have a ```@scenarios('<feature_file>`)``` decorator which can generate all scenarios in specific feature file?

Then it could look like:
```
@scenarios("my.feature")
def _(fixture_for_all_scenarios_in_that_feature):
    pass
```
What do you think?

bubenkoff:
i think that can be done, but with a warning that you're limiting yourself
and it can look like:

scenarios('my.feature')

so not a decorated variant, but cumulative

spinus:
For my any solution which limits lines of code which I don't need is good :-)
+1 if it's possible to implement.

bubenkoff:
this is not going to happen, discussed in #89

PR #106, opened 2015-03-18
Implemented  shortcut to automatically bind scenarios to tests.

### LINKED ISSUE #103: Auto-create scenario test functions
pytest-bdd would be much more awesome if it would auto create scenario functions, so you wouldn't have to define empty functions just to make the test go. I think this either could be implemented as a test collector (also would make the test reporter more user friendly as it would show the feature file as the source of tests) or it could be just a invocation form of scenario function like:

scenario('feature/something.feature', auto=True)

bubenkoff:
closes #103, #89, #92, #90

### DIFF
--- a/pytest_bdd/scenario.py
+++ b/pytest_bdd/scenario.py
@@ -13,6 +13,7 @@ test_publish_article = scenario(
 import collections
 import inspect
 import os
+import re
 
 import pytest
 from _pytest import python
@@ -24,6 +25,7 @@ from .feature import (
     Feature,
     force_encode,
     force_unicode,
+    get_features,
 )
 from .steps import (
     execute,
@@ -33,10 +35,13 @@ from .steps import (
 )
 from .types import GIVEN
 
-
 if six.PY3:
     import runpy
     execfile = runpy.run_path
+
+
+PYTHON_REPLACE_REGEX = re.compile("\W")
+ALPHA_REGEX = re.compile("^\d+_*")
 
 
 def _inject_fixture(request, arg, value):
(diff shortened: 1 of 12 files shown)


pytest_bdd/scenario.py
```diff
@@ -363,3 +368,40 @@ def scenario(feature_name, scenario_name, encoding="utf-8", example_converters=N
...
+        (attr.__scenario__.feature.filename, attr.__scenario__.name)
+        for name, attr in module.__dict__.items() if hasattr(attr, '__scenario__'))
+    for feature in get_features(abs_feature_paths):
+        for scenario_name, scenario_object in feature.scenarios.items():
+            # skip already bound scenarios
+            if (scenario_object.feature.filename, scenario_name) not in module_scenarios:
+                @scenario(feature.filename, scenario_name)
+                def _scenario():
+                    pass
+                module.__dict__['test_' + make_python_name(scenario_name)] = _scenario
```
olegpidsadnyi: incremental index in case already exists

tests/feature/test_scenarios.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_scenarios(testdir):
+    """Test scenarios shortcut."""
+    testdir.makeconftest("""
+        import pytest
+        from pytest_bdd import given
+
+        @given('I have a bar')
+        def i_have_bar():
+            print('bar!')
```
olegpidsadnyi: assert?

blueyed: This seems to be done via the outer test (looking at the output).
Or is there something specific to assert in the inner test?

bubenkoff: i missed the assert of the printed output - just forgot

On 18 March 2015 at 16:07, Daniel Hahler <notifications@github.com> wrote:

> In tests/feature/test_scenarios.py
> <https://github.com/pytest-dev/pytest-bdd/pull/106#discussion_r26671227>:
>
> > @@ -0,0 +1,59 @@
> > +"""Test scenarios shortcut."""
> > +import textwrap
> > +
> > +
> > +def test_scenarios(testdir):
> > +    """Test scenarios shortcut."""
> > +    testdir.makeconftest("""
> > +        import pytest
> > +        from pytest_bdd import given
> > +
> > +        @given('I have a bar')
> > +        def i_have_bar():
> > +            print('bar!')
>
> This seems to be done via the outer test (looking at the output).
> Or is there something specific to assert in the inner test?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pytest-dev/pytest-bdd/pull/106/files#r26671227>.
>



-- 
Anatoly Bubenkov

bubenkoff: done

pytest_bdd/scenario.py
bubenkoff: done

PR #110, opened 2015-04-28
Added hook before the step is executed with evaluated parameters

olegpidsadnyi:
Currently there is a hook similar to pytest item_call is missing. It is called before the step is executed when all the depending fixtures are evaluated.

### DIFF
--- a/pytest_bdd/hooks.py
+++ b/pytest_bdd/hooks.py
@@ -10,6 +10,10 @@ def pytest_bdd_after_scenario(request, feature, scenario):
 
 
 def pytest_bdd_before_step(request, feature, scenario, step, step_func):
+    """Called before step function is set up."""
+
+
+def pytest_bdd_before_step_call(request, feature, scenario, step, step_func, step_func_args):
     """Called before step function is executed."""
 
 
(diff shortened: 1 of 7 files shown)


pytest_bdd/hooks.py
```diff
@@ -10,7 +10,11 @@ def pytest_bdd_after_scenario(request, feature, scenario):
 
 
 def pytest_bdd_before_step(request, feature, scenario, step, step_func):
-    """Called before step function is executed."""
+    """Called before step function is set up."""
+
+
+def pytest_bdd_before_step_call(request, feature, scenario, step, step_func, step_func_args):
+    """Called before step function is set up."""
```
bubenkoff: before step function is called

tests/feature/test_steps.py
```diff
@@ -177,12 +180,15 @@ def test_when_step_validation_error():
     reprec = testdir.inline_run("-k test_when_step_validation_error")
     assert reprec.ret == 1
 
-    reprec = testdir.inline_run("-k test_when_dependency_fails", '-vv')
+    reprec = testdir.inline_run("-k test_when_dependency_fails", '-vvl')
```
bubenkoff: no need

tox.ini
```diff
@@ -25,4 +25,4 @@ deps =
 
 [pytest]
 pep8maxlinelength=120
-addopts=-vvl
+addopts=-vv
```
bubenkoff: why would you change this?

PR #111, opened 2015-05-15
Fix incompatibility with xdist

mdevlamynck:
Apparently xdist needs the test discovery to return a consistently ordered test list from each worker.
The scenarios() helper doesn't always return test in the same order so test discovery fails.

I added a sort in the scenarios function. I'm kinda new to python so I don't know if there's a faster / better / more pythonic way to do it but at least it works  (maybe we can use a sorted list to make things faster ?).

### DIFF
--- a/pytest_bdd/scenario.py
+++ b/pytest_bdd/scenario.py
@@ -410,7 +410,9 @@ def scenarios(*feature_paths):
     module_scenarios = frozenset(
         (attr.__scenario__.feature.filename, attr.__scenario__.name)
         for name, attr in module.__dict__.items() if hasattr(attr, '__scenario__'))
-    for feature in get_features(abs_feature_paths):
+    feature_list = get_features(abs_feature_paths)
+    feature_list.sort()
+    for feature in feature_list:
         for scenario_name, scenario_object in feature.scenarios.items():
             # skip already bound scenarios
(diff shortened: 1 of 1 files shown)


pytest_bdd/scenario.py
```diff
@@ -410,7 +410,9 @@ def scenarios(*feature_paths):
     module_scenarios = frozenset(
         (attr.__scenario__.feature.filename, attr.__scenario__.name)
         for name, attr in module.__dict__.items() if hasattr(attr, '__scenario__'))
-    for feature in get_features(abs_feature_paths):
+    feature_list = get_features(abs_feature_paths)
```
olegpidsadnyi: features are sorted inside get_features() function

olegpidsadnyi:
We are using it extensively with xdist and didn't have such a failure. And there is sorting by feature name or filename inside get_features function. Could you make more detailed research? What fails, why order differs. Maybe something wrong with the location or the way the feature filename is determined?

bubenkoff:
@mdevlamynck still interested in this?
im about to close it

mdevlamynck:
I didn't have time to look into this since then. You can close it, I'll try to fix my setup and come back if there I find something interesting for this tool.

PR #115, opened 2015-06-04
Make FeatureError.__str__ compatible with Python 3

lrowe:
Without this tracebacks omit the FeatureError message, line and file.


bubenkoff:
thanks for noticing, 2.9.1 with the fix is out

PR #119, opened 2015-06-07
Allow any step ordering

### LINKED ISSUE #118: Allowing when steps to follow then steps?
While it is generally good practice to order steps Given/When/Then it can sometimes be expedient to check conditions part way through a complex scenario (e.g. a multi-page form.)

I think the ordering checks in [feature.py](https://github.com/pytest-dev/pytest-bdd/blob/2.10.0/pytest_bdd/feature.py#L222) should be relaxed.

Cucumber [doesn't distinguish between step types](https://github.com/cucumber/cucumber/wiki/Given-When-Then) internally and does not enforce ordering. With behave one can register for all keywords using the ``step()`` decorator and ordering is not enforced.

lrowe:
This allows any Given/When/Then step ordering as in Cucumber. Closes #118.

### DIFF
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -36,18 +36,6 @@ from . import types
 from . import exceptions
 
 
-@six.python_2_unicode_compatible
-class FeatureError(Exception):
-
-    """Feature parse error."""
-
-    message = u"{0}.\nLine number: {1}.\nLine: {2}.\nFile: {3}"
-
-    def __str__(self):
-        """String representation."""
-        return self.message.format(*self.args)
-
-
 # Global features dictionary
 features = {}
 
@@ -218,20 +206,6 @@ class Feature(object):
                 if not clean_line:
                     continue
                 mode = get_step_type(clean_line) or mode
-
-                if mode == types.GIVEN and prev_mode not in (
-                        types.GIVEN, types.SCENARIO, types.SCENARIO_OUTLINE, types.BACKGROUND):
-                    raise FeatureError("Given steps must be the first in withing the Scenario",
-                                       line_number, clean_line, filename)
-
-                if mode == types.WHEN and prev_mode not in (
-                        types.SCENARIO, types.SCENARIO_OUTLINE, types.GIVEN, types.WHEN):
-                    raise FeatureError("When steps must be the first or follow Given steps",
-                                       line_number, clean_line, filename)
-
-                if not self.background and mode == types.THEN and prev_mode not in types.STEP_TYPES:
-                    raise FeatureError("Then steps must follow Given or When steps",
-                                       line_number, clean_line, filename)
 
                 if mode == types.FEATURE:
                     if prev_mode != types.FEATURE:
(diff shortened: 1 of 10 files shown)


bubenkoff:
well, allowing when-then-when is one thing
but allowing any order including given looks really bad - it means people will be allowed to write bad tests by definition

lrowe:
I think its reasonable to warn people about bad tests, but parse errors are really offputting to new users. My only real use case for allowing ``Given`` anywhere is when debugging and wanting to place a ``Then trace`` or ``Then screenshot`` step at any point and move it with a simple copy/paste.

bubenkoff:
if there's no real use-case (assuming that #121 is merged), there's no need to allow too much strange things
when-then-when-then is already quite questionable
I'd suggest to limit the relaxing to only `when-then-when-then`

lrowe:
Ease of debugging is a very important feature for a testing framework. Maintaining compatibility with other Gherkin implementations ensures developers can apply their experience from other tools.

I also think consistency makes things easier to explain: "It's considered good practice to order your scenarios Given/When/Then, but like Cucumber and other Gherkin testing tools pytest-bdd does not enforce any particular ordering."

bubenkoff:
ok, did you try this feature branch with your setup, does it work fully (with 2.11.0 merged) ?

lrowe:
Rebased on current master and added a test showing reverse given after when after then.

bubenkoff:
Please answer to my previos comment

lrowe:
@bubenkoff 
> did you try this feature branch with your setup

I've only ported one scenario to pytest-bdd so far, it works. There is now an explicit test for this feature and the pytest-bdd tests all pass.

> does it work fully (with 2.11.0 merged)

It is rebased on current master [f326f55]

bubenkoff:
i'd suggest you to continue with the conversion and use this branch for now
when you're sure you're happy with the results, we'll have at least a real proof of the 'feature' is useful

lrowe:
Using this I've been able to port all of my tests over from behave.

bubenkoff:
very good news!
and you still need to be able to have given after when, then?

lrowe:
I don't have any Given steps in my current project, but it is something I've seen in previous projects using Cucumber.

olegpidsadnyi:
Martin Flowler is speaking of sections http://martinfowler.com/bliki/GivenWhenThen.html. We also validate that the same given can't be evaluated twice since the result is cached in pytest request. So making given step is even more problematic and would be a moving away from Gherkin.
Thinking about all this drives me to a conclusion that adding an option to not perform a strict checking is not a very good thing.
Maybe structuring the parser and scenario executor with strategies it even better. So that an extension can be written on top of pytest-bdd.

lrowe:
Gherkin has a canonical implementation in Cucumber. Given after then works in Cucumber so allowing it is moving closer to Gherkin not further away.

bubenkoff:
just curious, how many scenarios which when-then-when-then you have, eg is it really a lot of work to convert to proper bdd?

lrowe:
@bubenkoff all of our tests have multiple when/then. Our gherkin tests are more specification testing than real BDD.

bubenkoff:
2.12.0 is out

PR #126, opened 2015-06-09
Mention step type on step definition not found errors and in code generation

### LINKED ISSUE #125: StepDefinitionNotFoundError should report Given/When/Then keyword
Step definitions are registered for specific keywords, so it would be useful to report which keyword was used in the failed lookup.
```
pytest_bdd.exceptions.StepDefinitionNotFoundError: Step definition is not found: "I should see an element with the css selector "table.sticky-area > tbody > tr"". Line 6 in scenario "Table" in the feature "/Users/lrowe/encoded/src/encoded/tests/features/experiments.feature
```

bubenkoff:
closes #125

### DIFF
--- a/tests/generation/test_generate_missing.py
+++ b/tests/generation/test_generate_missing.py
@@ -39,13 +39,13 @@ def test_generate_missing(testdir):
 
     result.stdout.fnmatch_lines(
         [
-            'Step "I have a custom bar" is not defined in the scenario '
+            'Step Given "I have a custom bar" is not defined in the scenario '
             '"Code is generated for scenario steps which are not yet defined(implemented)" *',
         ]
     )
 
     result.stdout.fnmatch_lines(
-        ['Step "I have a foobar" is not defined in the background of the feature "Missing code generation" *']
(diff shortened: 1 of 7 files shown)


tests/generation/test_generate_missing.py
```diff
@@ -39,13 +39,13 @@ def test_missing_steps():
 
     result.stdout.fnmatch_lines(
         [
-            'Step "I have a custom bar" is not defined in the scenario '
+            'Step "Given I have a custom bar" is not defined in the scenario '
```
olegpidsadnyi: This step is defined in the python code as "I have a custom bar". It can also be used as "And/But I have a custom bar". But it will print out "Given I have.."... Isn't it better to format it differently to separate the actual name with quotes?
'Step Given "I have a custom bar" is not found' or something like this?

bubenkoff: yep, had that in mind, just Given step "I have" is not english at all
Step Given "I have is probably better

bubenkoff:
@olegpidsadnyi so i changed it to {type}:"{name}"

PR #130, opened 2015-06-15
Make it possible to relax strict Gherkin scenario validation

### LINKED ISSUE #118: Allowing when steps to follow then steps?

bubenkoff:
closes #118. closes #119

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.12.0
+------
+
+- Make it possible to relax strict Gherkin scenario validation (bubenkoff)
+
 2.11.3
 ------
 
--- a/README.rst
+++ b/README.rst
@@ -963,6 +963,31 @@ You can learn more about `functools.partial <http://docs.python.org/2/library/fu
 in the Python docs.
 
 
+Relax strict Gherkin language validation
+----------------------------------------
+
+If your scenarios are not written in `proper` Gherkin language, e.g. they are more like textual scripts, then
+you might find it hard to use `pytest-bdd` as by default it validates the order of step types (given-when-then).
+To relax that validation, just override a fixture `pytestbdd_strict_gherkin` to return `False`:
+
+test_publish_article.py:
+
+.. code-block:: python
+
+    import pytest
+    from pytest_bdd import scenario
+
+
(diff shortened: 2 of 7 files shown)


lrowe:
This doesn't work when the ``pytestbdd_strict_gherkin`` fixture is set in a conftest.py.

lrowe:
It also doesn't work with the ``scenarios`` helper when the fixture is in the test file.

lrowe:
Also needs to handle steps in background.

bubenkoff:
my god, so your background contains something else than given?

bubenkoff:
what's your code, please show the structure

lrowe:
Currently we have a feature like this:
```cucumber
Feature: Search
    Background:
        When I visit "/search"
        And I wait for the content to load

    Scenario: Search
        Then the title should contain the text "Search"

    Scenario: Search Antibodies
        When I click the link with text that contains "Data"
        And I click the link to "/search/?type=antibody_lot"
        And I wait for the content to load
        Then I should see at least 15 elements with the css selector "ul.nav.result-table > li"
        And I should see at least 5 elements with the css selector "div.box.facets > div.facet"
        ...
```

bubenkoff:
fixed in 2.12.1 please check

lrowe:
My tests still fail with 2.12.1 because the ``pytestbdd_strict_gherkin`` fixture is not respected when using ``scenarios``.

bubenkoff:
Right, sorry i forgot to fix that as well. Will do today

bubenkoff:
2.12.2 is out, please try

lrowe:
Thanks @bubenkoff. This is now working for me with the scenarios helper and the fixture in conftest.py.

PR #149, opened 2015-09-21
Fixes the hook registration since pytest 2.8. Requirement changed

olegpidsadnyi:
Method addhooks is deprecated since pytest2.8.
Why don't we just increase the version number to 2.16.0 with py.test 2.8 as a requirement.
Changing addhooks to add_hookspecs seems to help.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+\.DS_Store
 *.rej
 *.py[cod]
 /.env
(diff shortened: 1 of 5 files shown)


bubenkoff:
https://github.com/pytest-dev/pytest/issues/1062

The-Compiler:
I didn't see this before opening #163 - anyways, #163 fixes the same in a way which is backwards compatible to 2.7.

The-Compiler:
I think this can be closed now that #163 is merged, right?

The-Compiler:
Closing this as I think it's taken care of with #163.

Issue #162, opened 2015-11-17
AttributeError: '_HookCaller' object has no attribute 'spec_opts'


The-Compiler:
When running `tox`, I get lots of failures with the `py27-xdist` environment:

```
============================ FAILURES ============================
___________________________ test_steps ___________________________
[gw2] linux2 -- Python 2.7.10 /home/florian/proj/pytest-bdd/.tox/py27-xdist/bin/python2.7
request = <FixtureRequest for <Function 'test_steps'>>

    @scenario_args('Every step takes a parameter with the same name')
>   def test_steps():

request    = <FixtureRequest for <Function 'test_steps'>>

tests/args/parse/test_args.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pytest_bdd/scenario.py:151: in _execute_scenario
    scenario=scenario,
.tox/py27-xdist/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py:724: in __call__
    return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
.tox/py27-xdist/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py:338: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

hook = <_HookCaller 'pytest_bdd_before_scenario'>
methods = [<_pytest.vendored_packages.pluggy.HookImpl instance at 0x7f5eead5cf38>]
kwargs = {'feature': <pytest_bdd.feature.Feature object at 0x7f5eea586290>, 'request': <FixtureRequest for <Function 'test_steps'>>, 'scenario': <pytest_bdd.feature.Scenario object at 0x7f5eea586710>}

    self._inner_hookexec = lambda hook, methods, kwargs: \
>       _MultiCall(methods, kwargs, hook.spec_opts).execute()
E   AttributeError: '_HookCaller' object has no attribute 'spec_opts'

hook       = <_HookCaller 'pytest_bdd_before_scenario'>
[...]

PR #167, opened 2015-12-20
Disable xdist tests for now.

The-Compiler:
See #162.
Those can probably be re-enabled when pytest 2.9 is released, but for now let's
disable them to have useful Travis feedback.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 [tox]
 distshare={homedir}/.tox/distshare
-envlist=linters,py27,py27-xdist,py27-pytest-latest,py34
+#envlist=linters,py27,py27-xdist,py27-pytest-latest,py34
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -11,11 +12,11 @@ deps = -r{toxinidir}/requirements-testing.txt
 interpreter=python2.7
 commands={[testenv]commands} pytest_bdd --pep8
 
-[testenv:py27-xdist]
-basepython=python2.7
-commands=
-  py.test pytest_bdd tests -n3 --pep8 -rfsxX \
-        --junitxml={envlogdir}/junit-{envname}.xml
+#[testenv:py27-xdist]
```
olegpidsadnyi: Does it need to be commented out? It is just a section of configuration

PR #168, opened 2015-12-20
2.16.0 release

The-Compiler:
This is on top of #167 - I think that's all I have to do, other than tagging and pushing to PyPI, right?


olegpidsadnyi:
Yes, please tag and push to PyPi

Issue #169, opened 2016-01-14
@pytest.mark.scenario Does not appear to exist


jtpereyda:
## tl;dr
README references `@mark.scenario` decorator from `pytest.mark`, but it appears that plain old `@scenario` from `pytest` is the way to go.

## Problem
According to the readme, you can parameterize tests like this (selected excerpts from example):

    import pytest
    from pytest_bdd import mark, given, when, then

    ...
    @pytest.mark.parametrize(
        ['start', 'eat', 'left'],
        [(12, 5, 7)])
    @mark.scenario(
        'parametrized.feature',
        'Parametrized given, when, thens',
    )
    ...
    def test_parametrized(start, eat, left):
    ...

However, these import lines fail for me with the latest pytest-bdd:

    $ cat test_import_fail.py
    import pytest
    import pytest_bdd
    print("pytest: {0}", pytest.__version__)
    print("pytest_bdd: {0}", pytest_bdd.__version__)
    from pytest_bdd import mark, given, when, then

    $ ./env/bin/python test_import_fail.py
    pytest: 2.8.5
    pytest_bdd: 2.16.0
    Traceback (most recent call last):
      File "test_import_fail.py", line 5, in <module>
        from pytest_bdd import mark, given, when, then
    ImportError: cannot import name mark

## Solution
However, everything works fine for me if I change 

    from pytest_bdd import mark, given, when, then

to

    from pytest_bdd import scenario, given, when, then

and

    @mark.scenario(

to

    @scenario(

## Suggestion
Change README.

~~Also "parameterization" is misspelled "parametrization" in the README.~~ Apparently there are four ways to spell parametrise.

spinus:
Good catch

jtpereyda:
Closing since PR was merged

PR #170, opened 2016-01-14
Issue #169 fix : @scenario instead of @mark.scenario; added accompanying gherkin file, too.

jtpereyda:
Addressed Issue #169 (unless pytest_bdd really does have a mark.scenario).

In doing so, I tested out the code sample to make sure it really works. To do this, I needed to create the gherkin file, so I just threw it in to clear things up.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -621,14 +621,14 @@ The code will look like:
 .. code-block:: python
 
     import pytest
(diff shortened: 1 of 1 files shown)


spinus:
+1

The-Compiler:
LGTM, thanks!

Issue #139, opened 2015-07-23
Output should match the output of behave and lettuce


thedrow:
Currently the output does not match BDD style test runners output.
This is the only reason I chose why not to use this plugin.

```
py.test -vvvv
================================================================================================ test session starts ================================================================================================
platform linux2 -- Python 2.7.10 -- py-1.4.30 -- pytest-2.7.2 -- /home/omer/.virtualenvs/playground/bin/python2.7
rootdir: /home/omer/bddtest, inifile: 
plugins: bdd
collected 2 items 

test_outline.py::test_outlined[12-5.0-7] PASSED
test_outline.py::test_outlined[2-1.0-1] PASSED

============================================================================================= 2 passed in 0.01 seconds ==============================================================================================
```

This is taken from one your examples provided in the README.
I'd expect the plugin to report the steps themselves and their status.

olegpidsadnyi:
This is the standard output of the PyTest, however it supports pluggable reporting. We've implemented cucumber json format (BDD standard) that we use to integrate to Jenkins (See https://github.com/pytest-dev/pytest-bdd#reporting).
What exactly are you using lettuce/behave console output for? Maybe it can be solved with the cucumber json format?
You could also help writing lettuce-style reporting plugin if you like :)

thedrow:
I'm using it to show our product team that the features they specify work as intended.
The cucumber json format doesn't help us because we're using Travis and not Jenkins.
I don't think I have the time to contribute a new reporting plugin right now.

olegpidsadnyi:
I'll think about it. It shouldn't be too hard, because there are hooks that allow to log feature/scenario step execution and just dump it in the end of the pytest log. Do you have a link to some online example where the format is good for you?

thedrow:
Just run either lettuce or behave with their examples. Their output mostly match.

bubenkoff:
so please be kind to specify the output, as you're initiating a change

spinus:
@bubenkoff I'm working on this. I already have something working but not yet all features covered. State for now ![selection_002](https://cloud.githubusercontent.com/assets/950799/9423831/8eb4dbf8-48cc-11e5-90c6-ecf81c2d80ce.png)

Do you prefer it as PR or I should create separate plugin?

btw, I plan to do the same for HTML report.
PS. @thedrow I think the title of this issue could be like feature request rather then complaining about that others have this :-)

bubenkoff:
@spinus that looks very promising!
[...]

PR #171, opened 2016-01-24
Spinus issue 139 gherkin terminal reporter

olegpidsadnyi:
Still couple of tests are failing

### DIFF
--- a/pytest_bdd/gherkin_terminal_reporter.py
+++ b/pytest_bdd/gherkin_terminal_reporter.py
@@ -1,0 +1,99 @@
+# -*- encoding: utf-8 -*-
+import py
+import pytest
+
+from _pytest.terminal import TerminalReporter
+
+
+def add_options(parser):
+    group = parser.getgroup("terminal reporting", "reporting", after="general")
+    group._addoption(
+        '--gherkin-terminal-reporter',
+        action="store_true",
+        dest="gherkin_terminal_reporter",
+        default=False,
+        help=(
+            "enable gherkin output"
+        )
+    )
+
+
+@pytest.mark.trylast
+def pytest_configure(config):
+    if config.option.gherkin_terminal_reporter:
+        # Get the standard terminal reporter plugin and replace it with our
+        current_reporter = config.pluginmanager.getplugin('terminalreporter')
+        if current_reporter.__class__ != TerminalReporter:
+            raise Exception("gherkin-terminal-reporter is not compatibile with any other terminal reporter."
+                            "You can use only one terminal reporter."
+                            "Currently '{0}' is used."
+                            "Please decide to use one by deactivating {0} or gherkin-terminal-reporter."
+                            .format(current_reporter.__class__))
+        gherkin_reporter = GherkinTerminalReporter(config)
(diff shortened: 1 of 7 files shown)


olegpidsadnyi:
@spinus please pull and have a look. There are still couple of tests failing

olegpidsadnyi:
It passes with tox, however fails comparing regular and gherkin execution with "1 passed" vs "2 passed".

spinus:
Hey, thanks for this. I'll be fixing it tomorrow.

spinus:
@olegpidsadnyi I was looking at your comment, but not sure what do you mean.

olegpidsadnyi:
I don't know.. If I run this file alone as
py.test tests/feature/test_gherkin_terminal_reporter.py

it fails. passes CI though..

    @then("output must be formatted the same way as regular one")
    def output_must_be_the_same_as_regular_reporter(test_execution):
        reg = test_execution['regular']
        ghe = test_execution['gherkin']
        assert reg.ret == 0
        assert ghe.ret == 0
        #  last line can be different because of test execution time is printed
        reg_lines = reg.stdout.lines if reg.stdout.lines[-1] else reg.stdout.lines[:-2]
        reg_lines[-1] = re.sub(r' \d+\.\d+ ', ' X ', reg_lines[-1])
        ghe_lines = ghe.stdout.lines if ghe.stdout.lines[-1] else ghe.stdout.lines[:-2]
        ghe_lines[-1] = re.sub(r' \d+\.\d+ ', ' X ', ghe_lines[-1])
        for l1, l2 in zip(reg_lines, ghe_lines):
>           assert l1 == l2
E           assert '============...=============' == '=============...============='
E             - =========================== 1 passed in X seconds ===========================
E             ?                             ^
E             + =========================== 2 passed in X seconds ===========================
E             ?                             ^

spinus:
Oh ok, I see. I couldn't reproduce it locally as well that's why I was a little confused.
Can you paste your packages versions (I'll try to reproduce it)? Looks like one test with gherkin reported is called twice.

olegpidsadnyi:
I've put that line (self.stats.setdefault(cat, []).append(rep)) all the way down. Not sure if it is correct now, at least it doesn't duplicate the stats anymore and test is passing.

spinus:
Thank you for fixing that.

spinus:
As you wrote to me, I updated readme and changes file.

pytest_bdd/gherkin_terminal_reporter.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+    )
+
+
+def configure(config):
+    if config.option.gherkin_terminal_reporter:
+        # Get the standard terminal reporter plugin and replace it with our
+        current_reporter = config.pluginmanager.getplugin('terminalreporter')
+        if current_reporter.__class__ != TerminalReporter:
+            raise Exception("gherkin-terminal-reporter is not compatibile with any other terminal reporter."
```
The-Compiler: nitpick: compatibile -> compatible

olegpidsadnyi:
Fixed in #139

PR #172, opened 2016-01-24
Allow tags after background section - fixes #160.

### LINKED ISSUE #160: Tags cause Background to fail
```bash
venv/lib/python2.7/site-packages/pytest_bdd/feature.py:290: in __init__
    "Background section can only contain Given steps", line_number, clean_line, filename)
E   FeatureError: Background section can only contain Given steps.
E   Line number: 17.
E   Line: @wip.
```
```gherkin
Background:
  Given I am on the new sale page as an associate


@wip
Scenario: Create new item by adding it to a sale
```

As you can see above, the Background only contains Given steps but the tag (@wip) on the first Scenario is being incorrectly included in the Background and causing an error.

As a workaround I tried to turn off strict checking using the following code but I couldn't get that to work either:
```python
@fixture(scope='session')
def pytestbdd_strict_gherkin():
    return False
```

Thanks

The-Compiler:


spinus:
+1 nice

PR #176, opened 2016-03-18
Cleaned up hooks, fixed report serialization to use builtin types.

### LINKED ISSUE #175: Reporter.serialize is breaking inside execnet serialize while using xdist
Values that are coming from converters for the parametrization (including scenario example tables) can be non simple python types, but custom classes defined by a project.
Execnet doesn't know how to serialize them and breaks.

note: Take into account the refactoring for the hooks (configure/unconfigure) done in the PR #147

olegpidsadnyi:
Fixes #175

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+
+2.16.1
+------
+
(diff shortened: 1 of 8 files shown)


The-Compiler:
LGTM!

PR #185, opened 2016-06-21
Better error explanation for the steps defined outside of scenarios

### LINKED ISSUE #156: Wrong error with no scenarios
This is a minor thing, but I thought I'd report it as I noticed it when doing a minimal testcase for another issue.

With this (broken) feature file:

```
Given foo
When bar
Then baz
```

we get:

```
E   pytest_bdd.exceptions.FeatureError: Given steps must be the first within the Scenario.
E   Line number: 1.
E   Line: Given foo.
E   File: /home/florian/tmp/bdd/feat.feature
```

Instead I'd expect the same as with an empty feature file:

```
E   pytest_bdd.exceptions.NoScenariosFound: ['/home/florian/tmp/bdd/feat.feature']
```

olegpidsadnyi:
Fixes #156 
Actually the current error is correct, only the message is ambiguous.
It should be a FeatureError to communicate the line where it breaks and it should mention that the step is defined outside of a scenario or a background.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/185)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.16.2
+------
+
+- Fix FixtureDef signature for newer pytest versions (The-Compiler)
+- Better error explanation for the steps defined outside of scenarios (olegpidsadnyi)
+
 
 2.16.1
 ------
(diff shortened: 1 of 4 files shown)


The-Compiler:
LGTM! Travis is still running though...

The-Compiler:
Hmm, seems like the [test fails](https://travis-ci.org/pytest-dev/pytest-bdd/jobs/139165567#L314) on Travis?

olegpidsadnyi:
Now should be fine

Issue #123, opened 2015-06-08
Steps require careful indentation not to be interpreted as multiline


lrowe:
Many [Gherkin](https://github.com/cucumber/cucumber/wiki/Gherkin#gherkin-syntax) [examples](https://github.com/cucumber/cucumber/wiki/Given-When-Then#and-but) on the net indent And/But steps with an extra space:

``` cucumber
Scenario: Indent my steps
    When I indent with 4 spaces
     And I indent with 5 spaces to line up
```

This breaks due to pytest-bdd using indentation of multiline steps: https://pytest-bdd.readthedocs.org/en/latest/#multiline-steps

It also makes feature files extremely sensitive to files with mixed tabs/spaces, such as the [steps.feature](https://github.com/pytest-dev/pytest-bdd/blob/master/tests/feature/steps.feature) in the pytest-bdd tests.

bubenkoff:
not sure, it will require quotes or some other symbols to distinguish multiline steps, which is not nicer than 
having to stick to predefined format, which is better in many ways, i think

lrowe:
Cucumber/Behat/Behave all go with the docstring approach.

I think there's a lot of value in maintaining compatibility with the cucumber gherkin standard, but changing existing parsing rules is not something to be done lightly either.

It might be interesting to experiment with https://github.com/nathforge/gherkin-parser as it claims to be tested against all Cucumber and Behave test cases.

bubenkoff:
nice project indeed, i'll take a look

bubenkoff:
@lrowe so we've looked and that parser, and it seems to be ok-ish but needs lots of improvements, and there's no guarantee they will be accepted by the owner.
also to integrate it with pytest-bdd seems like a lot of work
are you sure it worth the effort?

lrowe:
I think the indentation sensitivity issue is worth addressing at some point, though I don't have a strong opinion on the best way of doing so.

Out of interest, what type of improvements do you think need to be made to that parser library?

bubenkoff:
not sure, first of all, it seems a overengineered in part of the parser, and at the same time doesn't have useful data structure to traverse in reverse direction - say from step to scenario and feature. classes would be much better than just dicts for that, and that's why we used them in pytest-bdd
but in general i agree it makes no sense to have multiple parsers of the same simple thing - gherkin language
i feel i'll work on the integration

pydolan:
[...]

PR #188, opened 2016-06-29
Allow spaces in tags

The-Compiler:
Even if I originally told @olegpidsadnyi I'd drop this, here I am :laughing: 

It'd be nice for things like `@xfail: Issue #123` and it also makes the code in `get_tags` a bit simpler IMHO.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/188)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-2.16.2
+2.17.0
 ------
 
 - Fix FixtureDef signature for newer pytest versions (The-Compiler)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Changelog
 - Fix FixtureDef signature for newer pytest versions (The-Compiler)
 - Better error explanation for the steps defined outside of scenarios (olegpidsadnyi)
 - Add a ``pytest_bdd_apply_tag`` hook to customize handling of tags (The-Compiler)
+- Allow spaces in tag names. This can be useful when using the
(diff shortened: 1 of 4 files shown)


olegpidsadnyi:
I don't mind. Syntax color highlight looks a bit weird to me when you have spaces.
By the way what about #? Isn't it a comment symbol?

The-Compiler:
> I don't mind. Syntax color highlight looks a bit weird to me when you have spaces.

Looks good in emacs/spacemacs - depends on the editor I guess :wink:

> By the way what about #? Isn't it a comment symbol?

Hm, good point - my example wouldn't work then, but you get what the use case is.

Not sure why Travis is failing, seems unrelated to my changes?

olegpidsadnyi:
In [13]: line = "      @scenario-passes"

In [14]: set((tag.lstrip('@') for tag in line.split(' @') if len(tag) > 1))
Out[14]: {'     ', 'scenario-passes'}

In [15]: set((tag[1:] for tag in line.split() if tag.startswith("@") and len(tag) > 1))
Out[15]: {'scenario-passes'}

line can be not unindented and contain leading spaces

The-Compiler:
Fixed `get_tags` with leading spaces and added some tests for it. `test_step_trace` still fails locally, I'll take a closer look.

The-Compiler:
Should be fixed now, things look good locally. Travis might still fail due to #189.

olegpidsadnyi:
Please rebase to squash commits, bump the version and add a line to changelog

The-Compiler:
Done - I plan to do a release once this and #189 is merged.

PR #189, opened 2016-06-29
Don't use skip-mark in test_apply_tag_hook

The-Compiler:
`@pytest.mark.skip` was only introduced in pytest 2.9. Use `xfail` for
testing instead which will work with older pytest versions as well.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/189)


olegpidsadnyi:
@The-Compiler could you have a look at resolving a conflict?

The-Compiler:
Fixed and rebased on latest master!

PR #200, opened 2016-10-20
Relay fixture results to recursive call of 'get_features'.

coddingtonbear:
# What this does & Why

Currently, were you to specify fixture settings that alter the behavior used when parsing tests (specifically, in my case, `pytestbdd_strict_gherkin`), those fixture settings are not passed along to the recursive call of `get_features`.  This adds that.


bubenkoff:
nice catch! thanks!

bubenkoff:
2.18.1 is out

PR #211, opened 2017-04-18
Fix #210 do not raise wrong Error for no strict gherkin feature

### LINKED ISSUE #210: Undocumented check for Background section
There is a patch about improving handling definition steps outside of sections: https://github.com/pytest-dev/pytest-bdd/commit/cc87b5b0ddfb4cff81b53ab323bc237e9452dcfe

Check in this patch leads to error for situation, when I try to create Background section in the following way:
Background:
     When ..........
     And ......
So using "And" Step is prohibited after "When" in "Background". (It works when I use "Given" instead of "When" in this section).

I can not find any explanation in documentation why it was changed. Could you please add some note about it in documentation?

skraynev:
The issue, that if user disable strict gherkin check, it will lead to
situation, when wrong Error be raised. For the current situation it was
a message, that Step is defined outside of Background section.

Current patch makes check for orphan steps more tolerant for features
without strict gherkin syntax.

Also README was updated to show explicitly, that using "When, Then" is
prohibited in "Background" section.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/211)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -884,6 +884,14 @@ steps, adding possibility to prepare some common setup for multiple scenarios in
 About background best practices, please read
 `here <https://github.com/cucumber/cucumber/wiki/Background#good-practices-for-using-background>`_.
 
+.. NOTE:: There is only step "Given" should be used in "Background" section,
+          steps "When" and "Then" are prohibited, because their purpose are
+          related to actions and consuming outcomes, that is conflict with
+          "Background" aim - prepare system for tests or "put the system
+          in a known state" as "Given" does it.
+          The statement above is applied for strict Gherkin mode, which is
+          enabled by default.
+
 
 Reusing fixtures
 ----------------
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -293,7 +293,11 @@ class Feature(object):
                     continue
                 mode = get_step_type(clean_line) or mode
 
-                if not scenario and prev_mode not in (types.BACKGROUND, types.GIVEN) and mode in types.STEP_TYPES:
+                allowed_prev_mode = (types.BACKGROUND, types.GIVEN)
+                if not strict_gherkin:
+                    allowed_prev_mode += (types.WHEN, )
+
+                if not scenario and prev_mode not in allowed_prev_mode  and mode in types.STEP_TYPES:
                     raise exceptions.FeatureError(
                         "Step definition outside of a Scenario or a Background", line_number, clean_line, filename)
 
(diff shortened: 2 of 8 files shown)


README.rst
```diff
@@ -884,6 +884,12 @@ steps, adding possibility to prepare some common setup for multiple scenarios in
 About background best practices, please read
 `here <https://github.com/cucumber/cucumber/wiki/Background#good-practices-for-using-background>`_.
 
+.. NOTE:: There is only step "Given" should be used in "Background" section,
```
bubenkoff: mention that it only applies to strict mode (default)

tests/feature/no_sctrict_gherkin.feature
```diff
@@ -0,0 +1,9 @@
+Feature: No strict Gherkin Background support
+
+    Background:
+        When foo has a value "bar"
+        And foo is not boolean
+        And foo has not a value "baz"
+        Then foo has length equal 1
```
bubenkoff: no test for the strict mode validation

bubenkoff: also including even THEN in the background blows my mind - can't imagine the real use-case and why should we even test for such a silly thing?

pytest_bdd/feature.py
```diff
@@ -293,7 +293,11 @@ def __init__(self, basedir, filename, encoding="utf-8", strict_gherkin=True):
                     continue
                 mode = get_step_type(clean_line) or mode
 
-                if not scenario and prev_mode not in (types.BACKGROUND, types.GIVEN) and mode in types.STEP_TYPES:
+                allowed_prev_mode = (types.BACKGROUND, types.GIVEN)
+                if not strict_gherkin:
+                    allowed_prev_mode += (types.WHEN, )
```
bubenkoff: this also releases the checks for normal scenarios without any background, is it indended?

README.rst
skraynev: ok

tests/feature/no_sctrict_gherkin.feature
skraynev: ok, I will remove "then". I have not aim to blow your mind ;)

pytest_bdd/feature.py
skraynev: Could you explain what do you mean here?
"normal scenarios" - with disabled strict_gherkin or something else?
I can add test for it, but please describe the case.

skraynev:
patch is updated

pytest_bdd/feature.py
bubenkoff: yes, scenarios without background with disabled strict_gherkin

bubenkoff:
@skraynev sorry forgot to mention: please add your name to authors and add line to changelog

pytest_bdd/feature.py
skraynev: Khm... Yes it's true, so if user write such tests (with disabled strict gherkin), it means, that he consciously release all these checks. Probably it can be done just for writing some specific features.

skraynev:
Guys, sorry for the delay. I updated patch according your last suggestions.

bubenkoff:
@skraynev 2.18.2 is out, thanks!

PR #220, opened 2017-07-07
Adds ability to pass custom template to the generate command

thaffenden:
The generate option can be a great time saver, but it would really help me if I could use a custom template not the default one, that way I can configure it to save me the most time possible.

I've added a test which uses a new custom template and made a note about using it in the docs, but let me know if you think this needs anything else.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1159,6 +1159,16 @@ It will print the generated code to the standard output so you can easily redire
 
     pytest-bdd generate features/some.feature > tests/functional/test_some.py
 
+If you want to use a custom template to generate the test file you can pass the ``--template`` parameter to the command.
+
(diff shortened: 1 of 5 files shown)


thaffenden:
Sorry to bump but is this something anyone would be able to take a look at?

ggens:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pytest-dev/pytest-bdd/220)*

Couperin100:
Very useful functionality for Pytest-BDD.  Looks like its been reviewed so can we get this merged?

PR #221, opened 2017-07-18
Move the step evaluation into a hook

vodik:
Putting this up to get some feedback and see how receptive people are of the idea before I spend much more time on polishing this.

I'm currently using this library to try and bring this style of testing to our regression, but lots of our code is heavily built around asyncio.

I want to have control over how steps are evaluated, so I can automatically jump in when there's a coroutine and evaluate the step automatically in the current event loop. I'm currently using something like this to demo pytest-bdd internally:

```python
@pytest.hookimpl
def pytest_bdd_after_step(request, feature, scenario, step, step_func, step_func_args):
    if not inspect.iscoroutinefunction(step_func):
        return

    loop = asyncio.get_event_loop()
    loop.run_until_complete(step_func(**step_func_args))
    return True
```

Don't expect this merged as is, as its is still missing unit tests, documentation, and some quality of life stuff to suppress pluggy from the traceback should a step fail.

Also not sure if you guys want to keep the before and after call hooks too, then, since technically they're now equivalent to having a hook wrapper around this new hook.

### DIFF
--- a/pytest_bdd/plugin.py
+++ b/pytest_bdd/plugin.py
@@ -70,6 +70,12 @@ def pytest_bdd_before_step(request, feature, scenario, step, step_func):
     reporting.before_step(request, feature, scenario, step, step_func)
 
 
+@pytest.mark.trylast
+def pytest_bdd_call_step(request, feature, scenario, step, step_func, step_func_args):
+    step_func(**step_func_args)
+    return True
+
+
 @pytest.mark.tryfirst
 def pytest_bdd_after_step(request, feature, scenario, step, step_func, step_func_args):
     reporting.after_step(request, feature, scenario, step, step_func, step_func_args)
--- a/pytest_bdd/hooks.py
+++ b/pytest_bdd/hooks.py
@@ -19,6 +19,11 @@ def pytest_bdd_before_step_call(request, feature, scenario, step, step_func, ste
     """Called before step function is executed."""
 
 
+@pytest.hookspec(firstresult=True)
+def pytest_bdd_call_step(request, feature, scenario, step, step_func, step_func_args):
+    """Call the underlying step."""
+
+
 def pytest_bdd_after_step(request, feature, scenario, step, step_func, step_func_args):
     """Called after step function is successfully executed."""
 
--- a/pytest_bdd/scenario.py
+++ b/pytest_bdd/scenario.py
@@ -134,7 +134,7 @@ def _execute_step_function(request, scenario, step, step_func):
 
         request.config.hook.pytest_bdd_before_step_call(**kw)
         # Execute the step.
-        step_func(**kwargs)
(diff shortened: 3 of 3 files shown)


anxuae:
Is it possible to integrate this feature in the next release?

ggens:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pytest-dev/pytest-bdd/221)*

ggens:
work very well for me . good job.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pytest-dev/pytest-bdd/221#-:-L0sgJ6-qfVAcg06DAnt:b-q1hf7h)*

vodik:
Okay, if that's the case, do you guys want to merge it in its current state? Is there documentation you might want me to update? Should I mark the before/after hooks as deprecated?

vodik:
Any chance we can get something like this in for the next release?

Victor-Savu:
I would also be interested in finding out the status for this PR. @bubenkoff and @olegpidsadnyi in your opinion, what would be the remaining steps to finish this pull request?

pjz:
Any idea when/if this will get merged?

pjz:
FWIW, I rebased this on master and used it with some asyncio steps.  It seemed to break on example tables, however, making all calls with the literal `<fieldname>` value instead of one of the values in the column.

jruizaranguren:
@bubenkoff, @youtux, @olegpidsadnyi, It seems that the only thing to merge this improvement is just to document the hook. Would it be merged if I complete that documentation?

pytest_bdd/plugin.py
```diff
@@ -70,6 +70,12 @@ def pytest_bdd_before_step(request, feature, scenario, step, step_func):
     reporting.before_step(request, feature, scenario, step, step_func)
 
 
+@pytest.mark.trylast
+def pytest_bdd_call_step(request, feature, scenario, step, step_func, step_func_args):
+    step_func(**step_func_args)
+    return True
```
youtux, line 76: I would probably return the result of the step_func
```suggestion
    return step_func(**step_func_args)
```

pytest_bdd/plugin.py
```diff
@@ -70,6 +70,12 @@ def pytest_bdd_before_step(request, feature, scenario, step, step_func):
     reporting.before_step(request, feature, scenario, step, step_func)
 
 
+@pytest.mark.trylast
```
youtux, line 73: not sure why this is a `trylast` rather than a `tryfirst`

youtux (commented):
Yes, we would probably merge this once there are tests and documentation added

PR #225, opened 2017-10-24
Use actual values in step names when reporting.

mjholtkamp:
Hi guys,

I noticed that example values are not filled in when reporting. I made some changes and I would like you to consider merging them.

This is my first time offering something to the pytest-bdd repo, so I might not be aware of all the rules/guidelines that you have.

Please let me know how you feel about it, and if I can change the PR to fit your project better.

Thanks in advance!

The commit message follows:
-------------------------------------
When reporting failed (or succeeded) tests, the
Example values are not reported in the result. For
example: 'there are <start> cucumbers' instead of
'there are 12 cucumbers'.

This makes debugging harder (imagine you are using
several user accounts and one account fails).

This change adds/updates a 'reporting_name'
attribute to the step which will be used as the
name when serializing.

Some considerations:
- The same step will have different values for
reporting name during the execution of the tests
since a scenario outline will have different
values for each example.
- The 'name' attribute is not overwritten since it
might create a backwards incompatible change.
- Another reason why it is not overwritten is
because we need the original to format the
subsequent runs (in case of scenario outlines)

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/225)


olegpidsadnyi:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[pytest_bdd/utils.py, line 80 at r1](https://reviewable.io:443/reviews/pytest-dev/pytest-bdd/225#-KxNm22WIwR6mx746GWm:-KxNm22WIwR6mx746GWn:bqz12og) ([raw file](https://github.com/pytest-dev/pytest-bdd/blob/7ff27f45d6d8b7fc25e33ff7b42262767a47ab9f/pytest_bdd/utils.py#L80)):*
> ```Python
>     name = step.name
>     for param in step.params:
>         name = name.replace('<{}>'.format(param), str(get_fixture_value(request, param)))
> ```

what if the test crashes exactly in the fixture?
what if the value is unicode? encode to utf-8 then?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pytest-dev/pytest-bdd/225)*

olegpidsadnyi:
I'm not sure about parameter substitution in the report. Isn't it the whole point that people would navigate to the step definition by finding it by the step definition (matching a string with <>)?
Pytest displays the parametrization table in the report. Maybe there's something else in that cucumber.json that is missing?
Do you have an example output and the desired one?

olegpidsadnyi:
Our report is basically cucumber json. https://relishapp.com/cucumber/cucumber/docs/formatters/json-output-formatter#scenario-outline
I would prefer to be compatible with the original cucumber design. Is there anything missing in our output?
Maybe you could use some custom cucumber.json formatter instead?
The whole point of the json is that you can have your formatter on the CI server to have the most preferable way of the output instead of modifying the original source.

mjholtkamp:
Thanks for the comments, they are good points. I'll make a new PR with changes to the cucumber reporting output, I missed that before.

Apparently it's missing the 'examples', so I'll try to add them.

PR #232, opened 2018-01-23
Fixes #201 and pytest compatibility.

### LINKED ISSUE #201: `But` keyword should be supported
Standard Gherkin (see https://cucumber.io/docs/reference) uses an additional keyword `But` that functions identically to `And`, but the contrasting nature of the word can often make scenarios slightly more readable. This is not currently supported by Pytest-bdd, but since `And` is already supported, I imagine supporting `But` as well would be trivial. This does however allow for better specifications.

Note: Cucumber also supports using `*` as a stand-in for `And` or `But`. It functions exactly the same way as both of those.

olegpidsadnyi:
1. Pytest 3.3.2 has removed request._fixture_values, so ignoring AttributeError should be a solution. I guess replacing a FixtureDef with the cached_result will do the inject_fixture job.
2. There were compatibility issues introduced since pytest 2.8, and there's a change in the names of the parametrized items which breaks pytest-bdd tests. Let's support 2.8.1 as a minimal version then.
3. But keyword support added.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/232)

### DIFF
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -52,8 +52,9 @@ STEP_PREFIXES = [
     ("When ", types.WHEN),
     ("Then ", types.THEN),
     ("@", types.TAG),
-    ("And ", None),  # Unknown step type,
-    ("But ", None),  # Unknown step type,
+    # Continuation of the previously mentioned step type
+    ("And ", None),
+    ("But ", None),
 ]
 
 STEP_PARAM_RE = re.compile("\<(.+?)\>")
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -53,6 +53,7 @@ STEP_PREFIXES = [
     ("Then ", types.THEN),
     ("@", types.TAG),
     ("And ", None),  # Unknown step type,
+    ("But ", None),  # Unknown step type,
 ]
 
 STEP_PARAM_RE = re.compile("\<(.+?)\>")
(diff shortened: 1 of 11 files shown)


pytest_bdd/feature.py
```diff
@@ -53,6 +53,7 @@
     ("Then ", types.THEN),
     ("@", types.TAG),
     ("And ", None),  # Unknown step type,
```
bubenkoff: and why is it unknown here btw

pytest_bdd/feature.py
```diff
@@ -53,6 +53,7 @@
...
+    ("But ", None),  # Unknown step type,
```
bubenkoff: why is it unknown

pytest_bdd/steps.py
```diff
@@ -304,16 +304,19 @@ def inject_fixture(request, arg, value):
     fd.cached_result = (value, 0, None)
 
     old_fd = get_request_fixture_defs(request).get(arg)
-    old_value = get_fixture_value_raw(request, arg)
     add_fixturename = arg not in request.fixturenames
 
+    old_value = get_fixture_value_raw(request, arg)  # Compatibility with pytest < 3.3.2
```
bubenkoff: may be it's just easier to support only >=3.3.2?

pytest_bdd/feature.py
olegpidsadnyi: Changing it into # Continuation of the previously mentioned step type

pytest_bdd/steps.py
olegpidsadnyi: The plan is to fix the build and later work on some issues in order to do a major release. So that others will be able to work on fixes as well.

PR #246, opened 2018-07-05
Mostly factor out `py` module

asottile:
Still used for `py.io.TerminalWriter` so I added it to `setup.py`

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/246)

### DIFF
--- a/pytest_bdd/scripts.py
+++ b/pytest_bdd/scripts.py
@@ -28,6 +28,9 @@ def migrate_tests_in_file(file_path):
             content = fd.read()
             new_content = MIGRATE_REGEX.sub(r"\n@scenario(\2)\ndef \1():\n    pass\n", content)
             if new_content != content:
+                # the regex above potentially causes the end of the file to
+                # have an extra newline
+                new_content = new_content.rstrip('\n') + '\n'
                 fd.seek(0)
                 fd.write(new_content)
                 print("migrated: {0}".format(file_path))
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ setup(
         "Mako",
         "parse",
         "parse_type",
+        "py",
         "pytest>=2.8.1",
         "six>=1.9.0",
     ],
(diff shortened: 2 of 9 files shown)


pytest_bdd/scripts.py
```diff
@@ -26,7 +26,7 @@ def migrate_tests_in_file(file_path):
     try:
         with open(file_path, "r+") as fd:
             content = fd.read()
-            new_content = MIGRATE_REGEX.sub(r"\n@scenario(\2)\ndef \1():\n    pass\n", content)
+            new_content = MIGRATE_REGEX.sub(r"\n@scenario(\2)\ndef \1():\n    pass", content)
```
asottile: I wasn't sure about this here, by adjusting the test to use `dedent` instead of `Source` it showed that `migrate` was adding an extra newline here (I wasn't sure if this was intentional or not but it was causing the file to end with `\n\n` when previously it did not)

Another alternative I could take here is: `contents = contents.rstrip('\n') + '\n'` to ensure that the file ends in a newline

let me know!

asottile: actually, now that I look at this -- this'll cause

```python
foo = scenario(...)
bar = scenario(...)
```

to become

```python
@scenario(...)
def foo():
    pass
@scenario(...)
def bar():
    pass
```

instead of

```python
@scenario(...)
def foo():
    pass

@scenario(...)
def bar():
    pass
```

I think I'll leave this regex and adjust the newlines at the end of the file

setup.py
```diff
@@ -75,6 +75,7 @@ def run_tests(self):
         "Mako",
         "parse",
         "parse_type",
+        "py",
```
bubenkoff: should we try to remove it then completely in this PR?

asottile: the only thing left is `py.io.TerminalWriter` -- which even `pytest` still uses :( (to be factored out eventually as `py` is phased out)

PR #248, opened 2018-07-10
gherkin_terminal_reporter: add step-specific markup

### LINKED ISSUE #214: --gherkin-terminal-reporter doesn't color steps correctly
The gherkin terminal reporter doesn't color the steps correctly.

Let's say I have the following feature with a broken step:

```
Scenario: Publishing the article
    Given I'm an author user
    And I have an article
    When I go to the article page  # This step is broken
    And I press the publish button
    Then I should not see the error message
    And the article should be published
```

I expect the output to look like this (the first word in each line is the colouring):
```
RED:   Scenario: Publishing the article
GREEN:     Given I'm an author user
GREEN:     And I have an article
RED:       When I go to the article page  # This step is broken
RED:       And I press the publish button
RED:       Then I should not see the error message
RED:       And the article should be published
RED:   FAILED
```
But instead I get an output with all lines in red.
Printing the steps after the failed one is optional.

veliakiner:
Fixes #214

### DIFF
--- a/pytest_bdd/gherkin_terminal_reporter.py
+++ b/pytest_bdd/gherkin_terminal_reporter.py
@@ -100,12 +100,13 @@ class GherkinTerminalReporter(TerminalReporter):
                 self._tw.write(report.scenario['name'], **scenario_markup)
                 self._tw.write('\n')
                 for step in report.scenario['steps']:
+                    step_markup = {'red': True} if step["failed"] else {'green': True}
                     if self.config.option.expand:
                         step_name = self._format_step_name(step['name'], **report.scenario['example_kwargs'])
                     else:
                         step_name = step['name']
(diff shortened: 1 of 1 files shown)


pytest_bdd/gherkin_terminal_reporter.py
```diff
@@ -100,12 +100,13 @@ def pytest_runtest_logreport(self, report):
                 self._tw.write(report.scenario['name'], **scenario_markup)
                 self._tw.write('\n')
                 for step in report.scenario['steps']:
+                    step_markup = {'red': True} if step["failed"] else {'green': True}
```
feuillemorte: This is not working for `skipped` tests because of `else` statement

hristiy4n:
Any updates on this one?

Issue #253, opened 2018-09-25
pytest 3.8 generates warnings about directly calling fixtures


okken:
This rather unfortunately confusing warning occurs with pytest 3.8:

```
============= warnings summary =================
<path>\venv\lib\site-packages\pytest_bdd\scenario.py:228:\
RemovedInPytest4Warning: Fixture "pytestbdd_feature_base_dir" called directly. \
Fixtures are not meant to be called directly, are created automatically when test functions\
request them as parameters. \
See https://docs.pytest.org/en/latest/fixture.html for more information.\
  return call_fixture(getattr(fixtures, fixture))

<path>\venv\lib\site-packages\pytest_bdd\scenario.py:228: \
RemovedInPytest4Warning: Fixture "pytestbdd_strict_gherkin" called directly. \
Fixtures are not meant to be called directly, are created automatically when test functions \
request them as parameters. \
See https://docs.pytest.org/en/latest/fixture.html for more information.
  return call_fixture(getattr(fixtures, fixture))
```

Simple test file for playing with pytest-bdd:

try.feature:
```
Feature: One

  Scenario: First scenario
    Given some state
    When something happens
    Then result should occur
```

test_try.py:
```
from pytest_bdd import scenario, given, when, then

@scenario('try.feature', 'First scenario')
def test_first_scenario():
    pass

@given("some state")
def some_state():
    pass

@when("something happens")
def something_happens():
    pass

@then("result should occur")
def result_should_occur():
    pass
```

youtux:
pytest_bdd is indeed calling a fixture function, and this behavior is now deprecated by pytest and it will be removed in pytest 4.
It will require a major rewrite of pytest_bdd, PRs are welcome.

The-Compiler:
This seemed to work:

```diff
commit a0d65434d730777f9d6477bd4b748b558bb9fdf6
Author: Florian Bruhin <me@the-compiler.org>
Date:   Fri Sep 7 20:54:43 2018 +0200

    Work around pytest's warning for calling fixture functions
    
    pytest 3.7.0 introduced a warning when fixture functions are called directly.
    We do want to do that without triggering the warning, so we need to unwrap it.

diff --git a/pytest_bdd/scenario.py b/pytest_bdd/scenario.py
index f5b141b..45084e9 100644
--- a/pytest_bdd/scenario.py
+++ b/pytest_bdd/scenario.py
[...]

Issue #250, opened 2018-08-06
Fixtures not found with pytest 3.7


The-Compiler:
After upgrading from pytest 3.6.4 to 3.7.1, I get this with (I think) pretty much every test:

```
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/fixtures.py:520: in _get_active_fixturedef
    return self._fixture_defs[argname]
E   KeyError: 'pytestbdd_given_I open data/backforward/1.txt'

During handling of the above exception, another exception occurred:
.tox/py37-pyqt511/lib/python3.7/site-packages/pytest_bdd/scenario.py:92: in _find_step_function
    return get_fixture_value(request, get_step_fixture_name(name, step.type, encoding))
.tox/py37-pyqt511/lib/python3.7/site-packages/pytest_bdd/utils.py:36: in get_fixture_value
    return getfixturevalue(name)
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/fixtures.py:509: in getfixturevalue
    return self._get_active_fixturedef(argname).cached_result[0]
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/fixtures.py:523: in _get_active_fixturedef
    fixturedef = self._getnextfixturedef(argname)
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/fixtures.py:382: in _getnextfixturedef
    raise FixtureLookupError(argname, self)
E   _pytest.fixtures.FixtureLookupError: ('pytestbdd_given_I open data/backforward/1.txt', <FixtureRequest for <Function 'test_going_back_in_a_new_tab_without_history'>>)

During handling of the above exception, another exception occurred:
tests/end2end/features/test_backforward_bdd.py:49: in _scenario
    ???
.tox/py37-pyqt511/lib/python3.7/site-packages/pytest_bdd/scenario.py:163: in _execute_scenario
    step_func = _find_step_function(request, step, scenario, encoding=encoding)
.tox/py37-pyqt511/lib/python3.7/site-packages/pytest_bdd/scenario.py:105: in _find_step_function
    feature=scenario.feature,
E   pytest_bdd.exceptions.StepDefinitionNotFoundError: Step definition is not found: Given "I open data/backforward/1.txt". Line 47 in scenario "Going back in a new tab without history" in the feature "/home/florian/proj/qutebrowser/git/tests/end2end/features/backforward.feature
```

The-Compiler:
[...]

PR #255, opened 2018-10-07
Do not call fixtures directly

sliwinski-milosz:
Fixes: https://github.com/pytest-dev/pytest-bdd/issues/253

I added two new configs:
- bdd_feature_base_dir
- bdd_strict_gherkin

It is compatibility breaking change as no longer feature base dir or strict gherkin can be set by `pytestbdd_feature_base_dir` and `pytestbdd_strict_gherkin` fixtures.

Please note ini config is accessed through `pytest.config` and not through `request.config` object.

Documentation hasn't been updated yet.

Tests pass on pytest 3.6.4.
Some tests fail on the latest pytest (3.8.2). I guess it is due to https://github.com/pytest-dev/pytest-bdd/issues/250

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/255)


sliwinski-milosz:
I just committed fix for https://github.com/pytest-dev/pytest-bdd/issues/250

Maybe it will help someone with better understanding of pytest-bdd internals to find a better solution. However at the moment all tests pass. I have tested it also on `pytest==3.6.4` and `pytest==3.1.0`

sliwinski-milosz:
Sorry I had to rebase my branch... Now it is ready for review.

youtux:
Thanks for your contribution, @sliwinski-milosz. I am going to check it today, tomorrow at most.

pytest_bdd/scenario.py
```diff
@@ -298,9 +277,9 @@ def scenario(feature_name, scenario_name, encoding="utf-8", example_converters=N
     caller_function = caller_function or get_caller_function()
 
     # Get the feature
-    base_path = get_fixture(caller_module, "pytestbdd_feature_base_dir")
-    strict_gherkin = get_fixture(caller_module, "pytestbdd_strict_gherkin")
-    feature = Feature.get_feature(base_path, feature_name, encoding=encoding, strict_gherkin=strict_gherkin)
+    base_dir = get_base_dir(caller_module)
```
youtux: With the previous implementation, we could have test modules that don't want to define a base_dir (i.e. defaulting to the directory of the test module), and other tests modules that want to have feature files in a different location.
With this implementation, the use case is no longer possible.
What we could do is to add a `base_dir` parameter to (which defaults to None) to the `scenario` and `scenarios` functions, so that we still have a way to solve this use case.

pytest_bdd/scenario.py
```diff
@@ -298,9 +277,9 @@ def scenario(feature_name, scenario_name, encoding="utf-8", example_converters=N
...
+    strict_gherkin = get_strict_gherkin()
```
youtux: Same thing applies here

pytest_bdd/scenario.py
```diff
@@ -330,6 +309,20 @@ def scenario(feature_name, scenario_name, encoding="utf-8", example_converters=N
     )
 
 
+def get_base_dir(caller_module):
+    default_base_dir = os.path.dirname(caller_module.__file__)
```
youtux: this will result in the test file, not the folder containing it.

pytest_bdd/scenario.py
```diff
@@ -330,6 +309,20 @@ def scenario(feature_name, scenario_name, encoding="utf-8", example_converters=N
...
+    return get_from_ini('bdd_feature_base_dir', default_base_dir)
+
+
+def get_strict_gherkin():
+    return get_from_ini('bdd_strict_gherkin', True)
+
+
+def get_from_ini(key, default):
```
youtux: I'm not sure if we need this function just to fallback to a default. I'd rather declare the default in the pytest.config.addini(), and handle specific fallback behavior only where needed (in this case, in the get_base_dir function).

tests/feature/test_wrong.py
```diff
@@ -38,7 +39,7 @@ def then_nevermind():
 )
 @pytest.mark.parametrize('strict_gherkin', [True, False])
 @pytest.mark.parametrize('multiple', [True, False])
-@mock.patch('pytest_bdd.fixtures.pytestbdd_strict_gherkin', autospec=True)
+@mock.patch.object(sys.modules['pytest_bdd.scenario'], 'get_strict_gherkin', return_value=True)
```
youtux: I never saw patching sys.module directly. Can't we instead use something like
`@mock.patch.object('pytest_bdd.scenario.get_strict_gherkin')` instead?

Actually, if we are going to implement `strict_gherkin` as a parameter for `scenario`, then we don't even need to patch.

pytest_bdd/scenario.py
```diff
@@ -59,10 +58,15 @@ def find_argumented_step_fixture_name(name, type_, fixturemanager, request=None)
     # happens to be that _arg2fixturedefs is changed during the iteration so we use a copy
     for fixturename, fixturedefs in list(fixturemanager._arg2fixturedefs.items()):
         for fixturedef in fixturedefs:
-            parser = getattr(fixturedef.func, "parser", None)
+            if fixturedef.func.__name__ == 'lazy_step_func':
```
youtux: what are we fixing here exactly?

sliwinski-milosz: It is a good question ;)

It fixes https://github.com/pytest-dev/pytest-bdd/issues/250. For some unknown reason on the latest pytest above `fixturedef.func` don't have `parser` attribute -> even that on `pytest==3.6.4` `parser` attribute is there.

Because of that difference, `pytest-bdd` was not able to find some fixtures and tests were failing.

My guess is that previous `pytest` is calling `fixturedef.func` before our `find_argumented_step_fixture_name` is called.

The solution I found is to call the `func` if it is `lazy_step_func`. That call assigns  `parser` attribute and all work as before.

I found that solution during debugging but I don't know the root cause of the issue.

pytest_bdd/scenario.py
sliwinski-milosz: I am going to try to implement your solution.

Do you know any valid use case for that additional feature? Or by having that feature we encourage anti-patterns here? ;-)

However I agree that we shouldn't silently remove features. If we want to remove it, we should deprecate it firstly as someone may use it.

pytest_bdd/scenario.py
sliwinski-milosz: There is `os.path.dirname` that wraps `caller_module.__file__`. It returns directory of the current test file. I think all is good here. Correct me if I am wrong.

pytest_bdd/scenario.py
sliwinski-milosz: For `bdd_strict_gherkin` I already declared the default of `addini` call.
I'll remove `get_from_ini` call for `strict_gherkin`.

tests/feature/test_wrong.py
sliwinski-milosz: That was something new to me as well. It is there because of having `__all__` in `__init__.py` file that exposes only `scenario` function and not module.
To be able to access `scenario` module and mock one of its function you have to access it through `sys.modules`.

https://stackoverflow.com/questions/32026987/mock-patch-with-init-py

As you said, we don't need to mock it if we can pass `strict_gherkin` as parameter.

pytest_bdd/scenario.py
youtux: I would leave this change out and make a second PR for that, to be honest. It would make this case easier to solve.

pytest_bdd/scenario.py
youtux: I don't know how people use this library, so I am just careful to not kill use cases. We can even try a lazy approach, and implement what I was suggesting only if somebody opens an issue.
But in my case, I really need to be able to specify `strict_gherking` on some specific scenarios (by having it as a parameter of `scenario{,s}`). So if we implement this, let's also implement base_dir, it won't be much more difficult.

sliwinski-milosz:
Thank you very much for great review!

I applied review fixes and I moved fix for getting the parser to the separate pull request.

Please note that without that fix, tests on latest `pytest` will fail.

pytest_bdd/scenario.py
sliwinski-milosz: As you suggested I have moved it to separate pull request.

The only problem is that without that fix tests start to fail. May need to merge fix locally to this branch to make sure that all is good.

pytest_bdd/scenario.py
sliwinski-milosz: Done

youtux:
Yes, I think it makes more sense to call it `features_base_dir` indeed. I am going to try the new code today on one of my projects.

youtux:
@sliwinski-milosz can you re-merge master into your branch now? Tests should be passing.

sliwinski-milosz:
@youtux Tests on the latest pytest will not pass without https://github.com/pytest-dev/pytest-bdd/pull/256

That is why previously I had only one pull request - to have them green.
You can merge #256 locally to this branch and test.

The-Compiler:
> Please note ini config is accessed through `pytest.config` and not through `request.config` object.

This now raises deprecation warnings as of pytest 4.1.0 (which I configured as errors):

```
[...]
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:308: in load_module
    six.exec_(co, mod.__dict__)
tests/end2end/features/test_backforward_bdd.py:21: in <module>
    bdd.scenarios('backforward.feature')
.tox/py37-pyqt511/src/pytest-bdd/pytest_bdd/scenario.py:357: in scenarios
    features_base_dir = get_features_base_dir(module)
.tox/py37-pyqt511/src/pytest-bdd/pytest_bdd/scenario.py:311: in get_features_base_dir
    return get_from_ini('bdd_features_base_dir', default_base_dir)
.tox/py37-pyqt511/src/pytest-bdd/pytest_bdd/scenario.py:319: in get_from_ini
    value = pytest.config.getini(key)
.tox/py37-pyqt511/lib/python3.7/site-packages/_pytest/main.py:177: in __getattr__
    warnings.warn(PYTEST_CONFIG_GLOBAL, stacklevel=2)
E   _pytest.warning_types.PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
```

sliwinski-milosz:
Yes but I think https://github.com/pytest-dev/pytest-bdd/pull/282 fixes that

PR #256, opened 2018-10-11
Fix getting the parser from the lazy step function on latest pytest

sliwinski-milosz:
It fixes https://github.com/pytest-dev/pytest-bdd/issues/250. For some unknown reason on the latest pytest `fixturedef.func` doesn't have `parser` attribute -> even that on `pytest==3.6.4` `parser` attribute is there.

Because of that difference, `pytest-bdd` was not able to find some fixtures.

My guess is that previous `pytest` calls `fixturedef.func` before our `find_argumented_step_fixture_name` is called.

The solution I found is to call the `func` if it is `lazy_step_func`. That call assigns  `parser` attribute and all work as before.

I found that solution during debugging but I don't know the root cause of the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/256)


sliwinski-milosz:
Fixed with https://github.com/pytest-dev/pytest-bdd/pull/257 using slightly different approach.

PR #258, opened 2018-10-18
Bump to version 3.0.0

youtux:
PR #255 introduced a backward incompatible change (there was no way out of that), so we have to bump to a major version.

---
This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/258)


olegpidsadnyi:
looks good

PR #259, opened 2018-10-18
Add configuration for stale bot

youtux:


The-Compiler:
I fail to see the point with <40 open issues (and not many new ones). It'd probably be an annoyance rather than a help, because it's easier for a human to verify whether the currently open issues are still valid (irrespective of their age) or not.

youtux:
I think ~40 open issues are already too many to go through and check one by one if they are still relevant. Would this hurt? I think it would help to keep the number of issues to a minimum.

The-Compiler:
I think it would hurt - in a relatively slow-moving project with a handful of contributors, it can take a while until there's someone to take care of feature requests or even bugs (which is fine, nobody is getting paid for this after all, afaik). 

Then you end up having to either re-open issues by hand again, or you get duplicates for issues which were already reported before but closed. Or you'll need to keep things in your head rather than having a todo list in form of issues, which is a step back in my opinion.

I won't block this as a I haven't been very active in pytest-bdd development in the last couple of months, and other people might have a different opinion of course.

youtux:
> I think it would hurt - in a relatively slow-moving project with a handful of contributors, it can take a while until there's someone to take care of feature requests or even bugs (which is fine, nobody is getting paid for this after all, afaik).
> Then you end up having to either re-open issues by hand again, or you get duplicates for issues which were already reported before but closed. Or you'll need to keep things in your head rather than having a todo list in form of issues, which is a step back in my opinion.

We can use labels for issues we want to implement a solution for (`pinned` for example). But if you think it will not be beneficial, let's close this, no hard feelings.

PR #260, opened 2018-10-25
Update lowest required pytest version

sliwinski-milosz:
We use `bool` type ini option for `bdd_strict_gherkin`.

Support for that type has been added to `Parser.addini` starting from `pytest==2.9.0`.
https://docs.pytest.org/en/latest/changelog.html#id625

Due to above `pytest==2.9.0` is the lowest version that latest `pytest-bdd` can work with.

For testing with `pytest==2.9.0` you have to use `pytest-xdist<=1.17` as newer `pytest-xdist` requires `pytest>=3.0.0`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+3.0.1
+------
+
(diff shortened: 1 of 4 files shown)


youtux:
Thank you again @sliwinski-milosz :)

Issue #261, opened 2018-10-25
RemovedInPytest4Warning: MarkInfo objects are deprecated


sliwinski-milosz:
```
/usr/local/lib/python3.5/site-packages/pytest_bdd/reporting.py:78: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  param_names = parametrize.args[0] if isinstance(parametrize.args[0], (tuple, list)) else [
```

Line that invokes warning:
https://github.com/pytest-dev/pytest-bdd/blob/master/pytest_bdd/reporting.py#L76

Replacing below (as suggested in warning message):
```
parametrize = node.keywords._markers.get('parametrize')
```
with:
```
node.get_marker('parametrize')
```
makes tests still pass and we get rid of warning however I need to dig deeper into reporting logic to understand if above fix don't break anything. `get_closest_marker` works too.

Tested on:
- `pytest-bdd==3.0.0`
- `pytest==3.9.2`

EDIT:
`get_marker` also triggers a warning as it is deprecated too. `get_closest_marker` does not trigger it.

youtux:
It seems to me that Node.get_marker is deprecated since pytest 3.6. I find it strange that using node.get_marker('parametrize') and then accessing parametrize.args doesn't trigger the warning.
I would just use get_closes_marker, to be honest. We have tests for reporting, so the feature should be covered.

sliwinski-milosz:
Sorry you are right, `get_marker` triggers warning too.
So `get_closes_marker` is a better solution here.

We still access use MarkInfo objects while testing pytest-bdd. Would be good to fix that as well:
```
/Users/milosz.sliwinski/workspace/pytest-bdd/tests/feature/test_outline.py:17: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
Please use node.get_closest_marker(name) or node.iter_markers(name).
Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
  assert test_outlined.parametrize.args == (
```

sliwinski-milosz:
Soon I'll push fix for this warning so please do not rush with `3.0.1` release as we could put this fix into it.

Vic152:
Hey Guys I still get this warning:

[...]

PR #262, opened 2018-10-28
Do not use deprecated MarkInfo objects on pytest>=3.6

sliwinski-milosz:
Fixes https://github.com/pytest-dev/pytest-bdd/issues/261

Please note that new `get_closest_marker_args` function returns the same args on all supported `pytest` versions.

I fixed 3 test cases however I do not know how to get rid of the warning in following one (if it is even possible):
`tests/feature/test_outline.py::test_empty_example_values`

### DIFF
--- a/pytest_bdd/utils.py
+++ b/pytest_bdd/utils.py
@@ -84,3 +84,27 @@ def get_request_fixture_names(request):
     Compatibility with pytest 3.0.
     """
     return request._pyfuncitem._fixtureinfo.names_closure
+
+
+def get_closest_marker_args(node, mark_name):
+    """In pytest 3.6 new API to access markers has been introduced and it deprecated
+    MarkInfo objects.
+
+    This function uses that API if it is available otherwise it uses MarkInfo objects.
+    """
+    try:
+        return get_closest_marker_args_using_get_closest_marker(node, mark_name)
+    except AttributeError:
+        return get_closest_marker_args_using_mark_objects(node, mark_name)
+
+
+def get_closest_marker_args_using_get_closest_marker(node, mark_name):
+    """Recommended on pytest>=3.6"""
+    marker = node.get_closest_marker(mark_name)
+    return marker.args if marker else None
+
+
+def get_closest_marker_args_using_mark_objects(node, mark_name):
+    """Deprecated on pytest>=3.6"""
+    marker = node.keywords._markers.get(mark_name)
+    return (marker.args[0], marker.args[1]) if marker else None
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,0 +1,19 @@
+def get_parametrize_markers_args(node):
+    mark_name = 'parametrize'
+    try:
+        return get_markers_args_using_iter_markers(node, mark_name)
+    except AttributeError:
+        return get_markers_args_using_mark_objects(node, mark_name)
+
+
+def get_markers_args_using_iter_markers(node, mark_name):
+    """Recommended on pytest>=3.6"""
+    args = []
+    for mark in node.iter_markers(mark_name):
+        args += mark.args
+    return tuple(args)
+
+
+def get_markers_args_using_mark_objects(node, mark_name):
+    """Deprecated on pytest>=3.6"""
+    return getattr(node.keywords._markers.get(mark_name), 'args', ())
--- a/pytest_bdd/reporting.py
+++ b/pytest_bdd/reporting.py
@@ -7,6 +7,7 @@ that enriches the pytest test reporting.
 import time
 
 from .feature import force_unicode
+from .utils import get_closest_marker_args
 
 
 class StepReport(object):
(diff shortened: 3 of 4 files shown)


tests/utils.py
```diff
@@ -0,0 +1,19 @@
+def get_parametrize_markers_args(node):
+    mark_name = 'parametrize'
+    try:
+        return get_markers_args_using_iter_markers(node, mark_name)
+    except AttributeError:
+        return get_markers_args_using_mark_objects(node, mark_name)
+
+
+def get_markers_args_using_iter_markers(node, mark_name):
```
youtux: I see now that for the tests you actually used node.iter_markers. Let's then also use it in pytest_bdd/utils.py then

pytest_bdd/utils.py
```diff
@@ -84,3 +84,27 @@ def get_request_fixture_names(request):
...
+    """
+    try:
+        return get_closest_marker_args_using_get_closest_marker(node, mark_name)
+    except AttributeError:
+        return get_closest_marker_args_using_mark_objects(node, mark_name)
+
+
+def get_closest_marker_args_using_get_closest_marker(node, mark_name):
+    """Recommended on pytest>=3.6"""
+    marker = node.get_closest_marker(mark_name)
```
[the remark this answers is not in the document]
sliwinski-milosz: I think that `closest_marker` would be good enough here as we use only `parametrize_args[0]` and `parametrize_args[1]` which are closest marker arguments names and their values.

However I changed my mind and I switched to `iter_markers` as I will need all makers arguments in other pull requests ;-)

pytest_bdd/utils.py
```diff
@@ -84,3 +84,27 @@ def get_request_fixture_names(request):
...
+    return marker.args if marker else None
+
+
+def get_closest_marker_args_using_mark_objects(node, mark_name):
+    """Deprecated on pytest>=3.6"""
+    marker = node.keywords._markers.get(mark_name)
```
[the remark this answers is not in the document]
sliwinski-milosz: Removed completely

sliwinski-milosz:
Thank you very much for review! I have taken closer look to the code. There are few issues there but I decided to go one by one and small steps. So for now let's just focus on getting rid of the warning message without making any other improvements.

I will fix remaining issues in other pull requests.

Thank you also for tests code snippet!

tests/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    """Recommended on pytest>=3.6"""
+    args = []
+    for mark in node.iter_markers(mark_name):
+        args += mark.args
+    return tuple(args)
+
+
+def get_markers_args_using_mark_objects(node, mark_name):
+    """Deprecated on pytest>=3.6"""
+    return getattr(node.keywords._markers.get(mark_name), 'args', ())
```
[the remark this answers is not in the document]
sliwinski-milosz: Done

tests/utils.py
sliwinski-milosz: Done

pytest_bdd/utils.py
```diff
@@ -84,3 +84,29 @@ def get_request_fixture_names(request):
...
+    mark_name = 'parametrize'
+    try:
+        return get_markers_args_using_iter_markers(node, mark_name)
+    except AttributeError:
+        return get_markers_args_using_get_marker(node, mark_name)
+
+
+def get_markers_args_using_iter_markers(node, mark_name):
+    """Recommended on pytest>=3.6"""
+    args = []
```
youtux:
```suggestion
    return [arg for mark in node.iter_markers(mark_name) for arg in mark.args]
```

pytest_bdd/utils.py
```diff
@@ -84,3 +84,29 @@ def get_request_fixture_names(request):
...
+    for mark in node.iter_markers(mark_name):
+        args += mark.args
+    return tuple(args)
+
+
+def get_markers_args_using_get_marker(node, mark_name):
+    """Deprecated on pytest>=3.6"""
+    return getattr(node.get_marker(mark_name), 'args', ())
```
youtux: why do we use getattr here? `node.get_marker(mark_name).args` should be fine, right?

sliwinski-milosz: `node.get_marker(mark_name)` may return `None` which does not have `args` so with above approach we may end up with `AttributeError`. That is why I used `getattr`. We could switch to:
```python
marker = node.get_marker(mark_name)
if marker:
    return marker.args
return ()
```
or something similar

youtux: right, didn't see that. No need to change.

pytest_bdd/utils.py
sliwinski-milosz: I usually don't use list comprehension for cases with two loops as in my opinion it harms readability. But if you really want I am happy to do the change.

We would also have to convert it to tuple here.

youtux: Is it really necessary to convert it to tuple? As you wish, it is just flavour.

sliwinski-milosz: `get_markers_args_using_get_marker` returns `tuple`. We don't have to convert it to the `tuple` but then sometimes we would return `list` and sometimes `tuple` by calling `get_parametrize_markers_args`.

But you are right. It is not necessary (tests are still passing if we return it as a `list`).

youtux: I see. Then no change is necessary.

Issue #247, opened 2018-07-10
--gherkin-terminal-reporter-expanded does nothing


veliakiner:
Regardless of whether `-s` or `-v` are set. `--gherkin-terminal-reporter` works fine.

feuillemorte:
feature:
```python
When I start a <client>
```
run pytest:
```bash
pytest --gherkin-terminal-reporter -vv
```
output:
```bash
When I start a <client>
```
run with expand option:
```bash
pytest --gherkin-terminal-reporter --gherkin-terminal-reporter-expanded -vv
```
output:
```bash
When I start a java
```

Seems, it's working fine. Maybe documentations doesn't explain this option fine...

Vic152:
Can you guys update the documentation I did not know how to do it until I came here.

jsa34:
gherkin-terminal-reporter-expanded is no longer a feature flag - it's the default

Issue #235, opened 2018-01-29
Step name doesn't contain argument value in scenario outline


andreypetrenko82:
When I use scenario outline, and I can see that step name doesn't contain argument value in report.
In debug mode we can see that ```step.name``` contains just a name of argument. And in ```step_func_args``` we can see a value of this argument. But generated report via ```--cucumberjson``` has just a step with argument name (not its value).
Example:
```
Then I select <user_role> role
    Examples: user roles
      | user_role  |
      | Custom     |
```
And as a result generated report should contain:
```Then I select Custom role```
But it contains just an argument name: ```Then I select <user_role> role```

**Comment by olegpidsadnyi** [comment]
@andreypetrenko82 is it related to the new pytest? could you submit the complete test?

**Comment by slawek367** [comment]
Hi,
i saw something similar for example when I start test with: py.test --gherkin-terminal-reporter functional/some_test.py

Output contain only keys from example table but not values, so if you implement example:
Examples:
| some_key |
| value1      |
| value2      |

You will se in output:
for value1: Given some <some_key> ...
for value2: Given some <some_key> ...

Instead:
Given some value1
Given some value2

It is clear now?

**Comment by vinhntb** [comment]
@olegpidsadnyi  It's same issue #202. 
My scenario:

`

Feature: This is example demo

    Scenario Outline: Outlined given, when, thens
        Given there are <start> cucumbers
        When I eat <eat> cucumbers
        Then I should have <left> cucumbers

        Examples:
        | start | eat | left |
        |  12   |  5  |  7   |
        |  5    |  4  |  1   |
`

And output:
![image](https://user-images.githubusercontent.com/7419227/37636134-b4c83bd6-2c32-11e8-8445-baaf69b64090.png)

**Comment by sliwinski-milosz** [comment]
For Cucumber format add `--cucumberjson-expanded` to your command:
```
py.test --cucumberjson=<path to json report> --cucumberjson-expanded
```

For Gherkin reporter add `--gherkin-terminal-reporter-expanded` to your command:
```
py.test -s -vvv --gherkin-terminal-reporter --gherkin-terminal-reporter-expanded
```

jsa34:
Features requested I believe are now implemented.  Feel free to reopen if the gherkin terminal reporter is missing anything that I have overlooked.  Expanded reporter no longer exists and the normal terminal reporter should provide the output requested above

PR #263, opened 2018-10-29
Let gherkin-terminal-reporter-expanded enable gherkin output

sliwinski-milosz:
Related to:
https://github.com/pytest-dev/pytest-bdd/issues/247
https://github.com/pytest-dev/pytest-bdd/issues/235

This pull request updates `--gherkin-terminal-reporter-expanded` to enable gherkin output.
Thanks to that it can be used without adding `--gherkin-terminal-reporter` to the command.

This is the way suggested in the docs:
`https://pytest-bdd.readthedocs.io/en/latest/#reporting`

### DIFF
--- a/pytest_bdd/cucumber_json.py
+++ b/pytest_bdd/cucumber_json.py
@@ -32,7 +32,7 @@ def add_options(parser):
         "--cucumberjson-expanded",
         "--cucumber-json-expanded",
         action="store_true",
-        dest="expand",
+        dest="cucumber_json_expanded",
         default=False,
         help="expand scenario outlines into scenarios and fill in the step names",
     )
@@ -42,7 +42,7 @@ def configure(config):
     cucumber_json_path = config.option.cucumber_json_path
     # prevent opening json log on slave nodes (xdist)
     if cucumber_json_path and not hasattr(config, "slaveinput"):
-        config._bddcucumberjson = LogBDDCucumberJSON(cucumber_json_path, expand=config.option.expand)
+        config._bddcucumberjson = LogBDDCucumberJSON(cucumber_json_path, expand=config.option.cucumber_json_expanded)
         config.pluginmanager.register(config._bddcucumberjson)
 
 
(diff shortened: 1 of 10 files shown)


youtux:
@sliwinski-milosz Can you please add some tests for this? In case you need some input on how to test pytest plugins like pytest_bdd, you may want to check https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins.

Also, thank you very much for this PR.

youtux:
Also, running `pytest   --gherkin-terminal-reporter-expanded  tests` on the pytest_bdd project itself results in an `INTERNALERROR`. Would you be interested to look into that?

sliwinski-milosz:
User can parametrize scenario by using examples specified in feature files. In that case pytest `parametrize` marker is updated with examples parameters.

We also provide a way to parametrize scenario on the python side by using `pytest.mark.parametrize`. https://pytest-bdd.readthedocs.io/en/latest/#combine-scenario-outline-and-pytest-parametrization 

The problem with gherkin reporting is that when we are getting `self.example_kwargs`, we are using `scenario.get_example_params()` method that takes into account only args that are defined inside feature file (examples), and skips `pytest.mark.parametrize` parameters.

For gherkin reporting we need all parameters: those specified in feature files and those specified by using `pytest.mark.parametrize` decorator.

This pull requests uses `parametrize` marker to get all args.
It is ready for review.

This pull request contains three changes:
1. It enables gherkin output by using only gherkin-terminal-reporter-expanded option
2. It fixes issue mentioned above -> it gets all args for `self.example_kwargs`.
3. It fixes issue with unicode step names (which caused INTERNALERROR during testing).

If that is too much I can split this pull request into three smaller ones.

sliwinski-milosz:
As this pull request got quite messy I decided to split it into three smaller pull requests to fix issues one by one:
1. https://github.com/pytest-dev/pytest-bdd/pull/275
2. https://github.com/pytest-dev/pytest-bdd/pull/272
3. https://github.com/pytest-dev/pytest-bdd/pull/274

PR #264, opened 2018-10-29
Typo fixed in docs: strict_gherking->strict_gherkin

sliwinski-milosz:


youtux:
Thanks @sliwinski-milosz

PR #265, opened 2018-11-01
Test all supported pytest versions (plus more python interpreters)

youtux:
Test all the supported pytest versions (and latest pytest on all the supported python interpreters) and re-enable pytest-xdist in tox.ini and on trevis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,18 @@ python: "2.7"
 matrix:
   include:
     - env: TESTENV=py27-pytestlatest-linters
-    - env: TESTENV=py27-pytest{29,30,31,32,33,34,35,36,37,38,39,latest}
+    - env: TESTENV=py27-pytest29
+    - env: TESTENV=py27-pytest30
+    - env: TESTENV=py27-pytest31
+    - env: TESTENV=py27-pytest32
+    - env: TESTENV=py27-pytest33
+    - env: TESTENV=py27-pytest34
+    - env: TESTENV=py27-pytest35
+    - env: TESTENV=py27-pytest36
+    - env: TESTENV=py27-pytest37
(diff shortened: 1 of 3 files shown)


youtux:
@bubenkoff I could even enable all the combinations between interpreters and pytest versions on travis (i.e. `py{27,34,35,36,37}-pytest{29,30,31,32,33,34,35,36,37,38,39,latest}`). Please let me know what you think.

youtux:
I don't really see the point in testing against the master (unreleased) version of pytest, since it may have bugs that are not solved yet, and that would fail our build.
The reason behind testing for non-latest pytest releases is that sometimes we break compatibility (see https://github.com/pytest-dev/pytest-bdd/pull/260 for example) and this change would catch that.
We also have to explicitly maintain different implementation for some specific functions according to the pytest version (see https://github.com/pytest-dev/pytest-bdd/pull/262 for example). Without this change, we would only be testing the behavior of pytest-bdd with pytest > 3.6, but we would not cover the behaviour for pytest < 3.6.

bubenkoff:
the reason for testing master is to catch problems with released versions sooner
but i agree it's not critical
regarding older pytest versions, why can't we just stay on the latest pytest?
breaking compartibility is ok as soon as correct version requirement is there for pytest
the thing is, it makes support of pytest-bdd so much more problematic if we keep supporting multiple pytest versions

youtux:
Sometimes it gets quite difficult to upgrade pytest dependency on a project that use many pytest plugins and many pytest internals. That's why IMO we should not assume that everybody is using always latest pytest.

bubenkoff:
@youtux i see, ok
regarding making a huge matrix of different pythons and pytests i think it would be an overkill
so your changes are fine with me

PR #267, opened 2018-11-21
Add pytests 3.10 and 4 to tox

youtux:


tox.ini
```diff
@@ -1,11 +1,12 @@
 [tox]
 distshare={homedir}/.tox/distshare
-envlist=py27-pytestlatest-linters,py27-pytest{29,30,31,32,33,34,35,36,37,38,39,latest},py{34,35,36,37}-pytestlatest,py27-pytestlatest-xdist
+envlist=py27-pytestlatest-linters,py27-pytest{29,30,31,32,33,34,35,36,37,38,39,310,40,latest},py{34,35,36,37}-pytestlatest,py27-pytestlatest-xdist
 skip_missing_interpreters = true
 
 [testenv]
 deps =
     pytestlatest: pytest
+    pytest310: pytest~=3.10.0
```
sliwinski-milosz: Please add below to `testenv`:
```
    pytest40: pytest~=4.0.0
```

youtux: right

The-Compiler:
I don't think it makes sense to add both - from [the announcement](https://mail.python.org/pipermail/pytest-dev/2018-November/004660.html):

> This version is near identical to 3.10 feature-wise, except now pytest's own deprecated features generate errors

youtux:
@The-Compiler I know, but want to test against both, since pytest 4 now fails on warnings, and anyway pytest can introduce different (unexpected) behavior on the 3.10 branch.

PR #282, opened 2018-12-08
Fixes #276: configure and unconfigure Feature class

### LINKED ISSUE #276: CI job "py27-pytestlatest-xdist" fails
The job is failing lately, and this can be reproduced (without using xdist) using this particular schedule:
```
$ tests/feature/test_no_sctrict_gherkin.py::test_background_no_strict_gherkin  tests/feature/test_wrong.py::test_verbose_output
```

It seems that the setup of `test_background_no_strict_gherkin` leaks into `test_verbose_output`. I think this is a pytest issue and I am trying to create a smaller file to reproduce the issue.

sliwinski-milosz:
To be able to unconfigure Feature class during test execution we need to access previous config. I added `ConfigStack` for that.

I used that stack also to avoid accessing deprecated `pytest.config` while getting `config.hook.pytest_bdd_apply_tag`.

I also think that `def scenarios(*feature_paths, **kwargs):` needs refactor.

### DIFF
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -143,7 +143,7 @@ def get_tags(line):
     )
 
 
-def get_features(paths, caller_module, **kwargs):
+def get_features(paths, **kwargs):
     """Get features for given paths.
 
     :param list paths: `list` of paths (file or dirs)
@@ -159,13 +159,12 @@ def get_features(paths, caller_module, **kwargs):
                 features.extend(
                     get_features(
                         glob2.iglob(op.join(path, "**", "*.feature")),
-                        caller_module,
                         **kwargs
                     )
                 )
             else:
                 base, name = op.split(path)
-                feature = Feature.get_feature(base, name, caller_module, **kwargs)
+                feature = Feature.get_feature(name, base, **kwargs)
                 features.append(feature)
     features.sort(key=lambda feature: feature.name or feature.filename)
     return features
@@ -406,7 +405,7 @@ class Feature(object):
             cls.strict_gherkin = None
 
     @classmethod
-    def get_feature(cls, base_dir, filename, caller_module, encoding="utf-8", strict_gherkin=True):
+    def get_feature(cls, filename, base_dir=None, caller_module=None, strict_gherkin=True, encoding="utf-8"):
         """Get a feature by the filename.
 
         :param str base_path: Base feature directory.
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -143,7 +143,7 @@ def get_tags(line):
     )
 
 
-def get_features(paths, **kwargs):
+def get_features(paths, caller_module, **kwargs):
     """Get features for given paths.
 
     :param list paths: `list` of paths (file or dirs)
@@ -159,12 +159,13 @@ def get_features(paths, **kwargs):
                 features.extend(
                     get_features(
                         glob2.iglob(op.join(path, "**", "*.feature")),
+                        caller_module,
                         **kwargs
                     )
                 )
             else:
                 base, name = op.split(path)
-                feature = Feature.get_feature(base, name, **kwargs)
+                feature = Feature.get_feature(base, name, caller_module, **kwargs)
                 features.append(feature)
     features.sort(key=lambda feature: feature.name or feature.filename)
     return features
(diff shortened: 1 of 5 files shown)


sliwinski-milosz:
Can we firstly merge this pull requests as it fixes CI tests? Please review :)

pytest_bdd/plugin.py
```diff
@@ -47,14 +49,19 @@ def add_bdd_ini(parser):
 @pytest.mark.trylast
 def pytest_configure(config):
     """Configure all subplugins."""
+    ConfigStack.add(config)
```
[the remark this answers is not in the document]
sliwinski-milosz: It is all here to fix tests that we run using `testdir` plugin.

Our problem is that:
1. When we start all tests it calls `pytest_configure`.
2. Then when it starts some test with `testdir` it calls `pytest_configure` again with new config.
3. After test (as in current code there is no `pytest_unconfigure`) it just runs another test.
4. As `pytest_configure` from point 2 changed pytest config - our tests start to fail as they expect default config (from point 1).

I added `pytest_unconfigure` hook so now after running test using `testdir` it can bring back previously used config.

So this pull request changes logic to:
1. When we start all tests it calls `pytest_configure` (we store config on stack).
2. Then when it starts some test with `testdir` it calls `pytest_configure` again with new config (again we store new config on stack).
3. After tests it calls `pytest_unconfigure` -> we bring back previous config from the stack.
4. We have previous config back and tests pass :)

pytest_bdd/utils.py
```diff
@@ -110,3 +110,22 @@ def get_markers_args_using_iter_markers(node, mark_name):
...
+
+    @classmethod
+    def add(cls, config):
+        cls._stack.append(config)
+
+    @classmethod
+    def get_previous(cls):
+        cls._stack.pop()
+        if cls._stack:
+            return cls._stack[-1]
```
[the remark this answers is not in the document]
sliwinski-milosz: Added

pytest_bdd/utils.py
```diff
@@ -110,3 +110,22 @@ def get_markers_args_using_iter_markers(node, mark_name):
...
+
+    @classmethod
+    def get_pytest_bdd_apply_tag_hook(cls):
+        return cls._stack[-1].hook.pytest_bdd_apply_tag
```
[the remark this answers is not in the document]
sliwinski-milosz: 👍 Done

sliwinski-milosz:
This pull request fixes CI tests

The-Compiler:
Would be great if someone else could take a look at this as well - I'm quite busy with exams coming up, so I don't trust my brain enough to review bigger things right now :wink:

pytest_bdd/utils.py
```diff
@@ -110,3 +110,23 @@ def get_markers_args_using_iter_markers(node, mark_name):
 def get_markers_args_using_get_marker(node, mark_name):
     """Deprecated on pytest>=3.6"""
     return getattr(node.get_marker(mark_name), 'args', ())
+
+
+class ConfigStack(object):
```
youtux: Do we really need to define an object to handle a stack? I'd rather use just a list and name it `config_stack`

pytest_bdd/utils.py
```diff
@@ -110,3 +110,23 @@ def get_markers_args_using_iter_markers(node, mark_name):
...
+
+    _stack = []
+
+    @classmethod
+    def add(cls, config):
+        cls._stack.append(config)
+
+    @classmethod
+    def get_previous(cls):
```
youtux: also because get_previous should not have side effects, but instead it removes the current element and return the previous one

pytest_bdd/feature.py
```diff
@@ -250,6 +250,8 @@ def __bool__(self):
 class Feature(object):
 
     """Feature."""
+    strict_gherkin = None
```
youtux: I don't think that the Feature class should contain configuration options like `strict_gherkin` or `base_dir`. Instead, callers of `Feature(...)` should consult the current config from the stack.

pytest_bdd/utils.py
sliwinski-milosz: Changed to list

pytest_bdd/utils.py
sliwinski-milosz: Removed

sliwinski-milosz:
Closing in favour of:
https://github.com/pytest-dev/pytest-bdd/pull/288

PR #287, opened 2019-01-10
In tests use -m instead of -k for running tests by marker expressions

sliwinski-milosz:
In tests we use `-k` for running tests by markers (tags). That approach works for `pytest<4.1.0`. Pytest documentation says nothing about ability to use `-k` for selecting markers.
https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests

Correct approach is to use `-m` for markers.

Latest `pytest==4.1.0` removes some legacy code related to markers. My guess is that it also removed ability to use `-k` for selecting markers:
(removal of `store_legacy_markinfo` function)
https://github.com/pytest-dev/pytest/commit/9f9f6ee48beba8bbf0911e458590aa67b45bd867#diff-b1e6724759432cb1b166981e45a4a25cL238

Please note that some tests are failing because of the warnings that break regex checks of tests results. Those warnings are fixed in https://github.com/pytest-dev/pytest-bdd/pull/282

This and mentioned above pull request together make our tests green again.

### DIFF
--- a/tests/feature/test_tags.py
+++ b/tests/feature/test_tags.py
@@ -15,13 +15,6 @@ def test_tags(request):
 
     assert test.__scenario__.tags == set(['scenario_tag_1', 'scenario_tag_2'])
     assert test.__scenario__.feature.tags == set(['feature_tag_1', 'feature_tag_2'])
-
-    assert getattr(test, 'scenario_tag_1')
-    assert getattr(test, 'scenario_tag_2')
-
-    assert getattr(test, 'feature_tag_1')
-    assert getattr(test, 'feature_tag_2')
-
     test(request)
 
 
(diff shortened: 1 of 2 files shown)


tests/feature/test_tags.py
```diff
@@ -15,13 +15,6 @@ def test():
 
     assert test.__scenario__.tags == set(['scenario_tag_1', 'scenario_tag_2'])
     assert test.__scenario__.feature.tags == set(['feature_tag_1', 'feature_tag_2'])
-
-    assert getattr(test, 'scenario_tag_1')
-    assert getattr(test, 'scenario_tag_2')
-
-    assert getattr(test, 'feature_tag_1')
-    assert getattr(test, 'feature_tag_2')
```
sliwinski-milosz: Above asserts fail on `pytest>=4.1.0` - but I think that we don't really need those attributes (tags) on `test` function as we don't use them. Correct me if I am wrong.

sliwinski-milosz:
I would do that other way round. I would merge this pull request (as it is only tests update and is quite obvious) and then I would update #282 to check if tests pass :)

youtux:
Then we need the help of @pytest-dev/pytest-bdd-admin, as I don't have superpowers to merge this PR.

sliwinski-milosz:
I thought you have :D Could you please  approve this PR to let them know that you are fine with changes?

sliwinski-milosz:
Other option is to merge this PR to #282 . What do you think?

youtux:
Also fine to me, since we are effectively trying to fix CI builds due to pytest changes.

sliwinski-milosz:
Merged to https://github.com/pytest-dev/pytest-bdd/pull/282 in the forked repository

sliwinski-milosz:
@youtux as you have now superpowers, could you please merge this to master? It would make #282 easier to review ;-)

youtux:
I still don't have superpowers, @sliwinski-milosz

The-Compiler:
Superpowers, I hear? :see_no_evil:

PR #290, opened 2019-02-17
Drop ancient pytest compatibility

youtux:
Removed compatibility with pytest < 3.3.2
The test suite still appears to work with pytest >= 3.0.0, but let's remove the declared compatibility until 3.3.2 anyway (it is more that 1 year old).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
         "parse",
         "parse_type",
         "py",
-        "pytest>=2.9.0",
+        "pytest>=3.0.0",
         "six>=1.9.0",
     ],
     # the following makes a plugin available to py.test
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -76,7 +76,7 @@ def run_tests(self):
         "parse",
         "parse_type",
         "py",
-        "pytest>=2.9.0",
+        "pytest>=3.0.0",
```
sliwinski-milosz: 3.3.2

youtux: Right. But since the test suite was not failing on pytest >=3.0.0, I would say that we can leave the `pytest>=3.0.0` requirement, and if somebody reports an issue, we can definitively bump it to `pytest>=3.3.2`

PR #291, opened 2019-02-19
Treat pytest errors as warnings and fix those errors

youtux:


youtux:
the builds are failing just because coveralls is having troubles.

PR #294, opened 2019-03-12
Update outdated behat hyperlinks in documentation

sliwinski-milosz:
Instead of pointing to specific version (`v3.0` in this case) we could also consider using `latest` for example:
`http://behat.org/en/latest/user_guide/writing_scenarios.html#pystrings`
On the other hand hardcoding the version is safer.

Please note also that below url:
```
http://docs.behat.org/en/v2.5/guides/1.gherkin.html#backgrounds
```
is working fine but I updated it to `v3.0` to be consistent with other urls.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -379,7 +379,7 @@ Multiline steps
 ---------------
 
 As Gherkin, pytest-bdd supports multiline steps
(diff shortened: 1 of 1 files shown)


youtux:
Thanks Milosz!

PR #302, opened 2019-05-25
Use normal parameter substitution with scenario outlines

nicois:
Currently, scenario outlines use an alternate parser to the scenarios, resulting in needing to decorate step functions multiples times, and requiring gherkin authors to use identical parameters in scenario outlines to those used in the step definitions.

This patch will attempt to 'promote' examples to being fully-fledged scenarios, which the normal parsers can then map to individual step functions.

If example_converters have been defined (which was introduced as a hack to work around the fact the normal parsers were not compatible with scenario outlines), the legacy behaviour is maintained.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -22,5 +22,6 @@ These people have contributed to `pytest-bdd`, in alphabetical order:
 * `Leonardo Santagada <santagada@github.com>`_
 * `Milosz Sliwinski <sliwinski-milosz>`_
 * `Michiel Holtkamp <github@elfstone.nl>`_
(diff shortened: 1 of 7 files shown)


nicois:
I believe this is backwards-compatible, based on the regression tests passing.

However, I believe that it doesn't make sense to use the 'scenario' decorator with a scenario outline. In my mind, it makes more sense for each variant in a scenario outline to be treated as an independent test, which is how I have approached it in this PR.

Let me know if you are happy with this approach.

kotavi:
Is someone still working on this change?

elchupanebrej:
@youtux this could be closed after release 5.0

youtux:
Indeed

PR #306, opened 2019-06-28
Add gettext support

dutradda:
First translation was written for portuguese, according to https://cucumber.io/docs/gherkin/reference/#spoken-languages.

The PR also enables the language reset during runtime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ python:
   - "3.6"
   - "3.7"
 
-install: pip install tox tox-travis
+install:
+  - make translate
+  - pip install tox tox-travis
 
 script: tox --recreate
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ include *.rst
 include *.txt
 include setup.py
 include pytest_bdd/templates/*.mak
+include pytest_bdd/locales/*/LC_MESSAGES/*.mo
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,13 @@ PATH := .env/bin:$(PATH)
 develop: .env
 	pip install -e . -r requirements-testing.txt tox python-coveralls
 
-coverage: develop
+translate:
+	msgfmt pytest_bdd/locales/en/LC_MESSAGES/step-prefixes.po \
+		-o pytest_bdd/locales/en/LC_MESSAGES/step-prefixes.mo
+	msgfmt pytest_bdd/locales/pt_BR/LC_MESSAGES/step-prefixes.po \
+		-o pytest_bdd/locales/pt_BR/LC_MESSAGES/step-prefixes.mo
+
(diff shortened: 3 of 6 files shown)


dutradda:
@youtux, @olegpidsadnyi and @sliwinski-milosz 

Hello guys,
can any of you take a look on my PR?

I have to change the tox/travis settings because the pytest 4.5 and after are crashing the tests.

sliwinski-milosz:
Travis has been fixed on master so you can revert your travis related changes

dutradda:
@sliwinski-milosz 
code updated

olegpidsadnyi:
To be honest I don't really understand why do we need gettext for 11 keywords. This seems like an overkill. Another point is the current language. I think this should be a pytest option that the plugin introduces.

olegpidsadnyi:
The whole point is to be compatible with the cucumber? To let people use pytetst instead of cucumber? Then the language support is like this:
https://cucumber.io/docs/gherkin/reference/#spoken-languages

#language header, and only keywords. There's no need in gettext. This is just a simple dict lookup once the header is set on the feature file. Otherwise english.

See https://github.com/cucumber/gherkin-python/blob/master/gherkin/gherkin-languages.json

youtux:
Maybe we should instead import and use [gherkin](https://pypi.org/project/gherkin-official) and delegate the parsing of feature files to this library, provided by cucumber.io.
If we use that, we will have automatic support for all the languages, and issues like #311 would probably not exist, or they would have to be fixed by the upstream.

dutradda:
@olegpidsadnyi and @youtux 
How we can proceed?

I really want to see multiple languages support on `pytest-bdd`. We already written the feature files in portuguese here. And the Gherkin language recommends that.

I can change my PR to fit the needs of the project.
But I need to know the next steps.

youtux:
@dutradda, if you are willing to work on this, I would propose to continue in this way:

1. add the dependency to `gherkin-official` to this project
2. parse feature files with that library, instead of having our own parsing logic

I am not sure how much work it would require, but I think it is a viable option.

dutradda:
@youtux 
I'll do it and keep you informed

jsa34:
@youtux is this still relevant/needed?

youtux:
> @youtux is this still relevant/needed?

Now that we use the official parser it should come out of the box. https://github.com/pytest-dev/pytest-bdd/pull/698


maybe we should add a test though to make sure that everything work fine when using a feature file in  a different language

jsa34:
Good point.  I've got a few other issues that just need asserting and tests to close them.  I'll add this to them

jsa34:
@youtux I don't think this is needed if we are using the official parser and should probably use the gherkin languages.json as a lookup for gherkin syntax?

jsa34:
Localisation now supported.  This PR has been superseded.

PR #307, opened 2019-07-04
Fix CI suite problems

youtux:
I removed some old pytest version from the matrix, as well as python 3.4.


tests/feature/test_cucumber_json.py
```diff
@@ -31,6 +31,14 @@ def __cmp__(self, other):
 
 def test_step_trace(testdir):
     """Test step trace."""
+    testdir.makefile(".ini", pytest=textwrap.dedent("""
+    [pytest]
+    markers =
+        scenario-passing-tag
+        scenario-failing-tag
+        scenario-outline-passing-tag
+        feature-tag
```
sliwinski-milosz: Why do we need those markers here? Is it documented somewhere?

sliwinski-milosz:
Now we have 2 checks - it runs twice in the travis. What is the difference between those checks? :)

tests/feature/test_cucumber_json.py
youtux: They are used by the feature file of line 42. Pytest 4.5+ emits warnings if you se markers without defining them in pytest.ini, and we treat warnings as errors in our test configuration.

tests/feature/test_tags.py
```diff
@@ -1,25 +1,23 @@
...
-    @scenario(
-        'tags.feature',
-        'Tags'
-    )
-    def test():
-        pass
-
-    assert test.__scenario__.tags == set(['scenario_tag_1', 'scenario_tag_2'])
-    assert test.__scenario__.feature.tags == set(['feature_tag_1', 'feature_tag_2'])
-    test(request)
```
sliwinski-milosz: One last question... are you sure this test is not needed? ;-)

youtux: I don't it's purpose. We already have other tests that assert that pytest "sees" the tags, by asking pytest to run only certain items.

We don't expose any API for test.__scenario__, so what are we testing it for?
Moreover, this test was calling `test(request)`, what is even the point of that?

PR #312, opened 2019-07-25
Allow for scenario descriptions to be present

### LINKED ISSUE #311: Scenario descriptions cause odd results
Perhaps it is not part of pytest-bdd's "subset" of Gherkin, but adding a scenario description causes issues.

Here's a scenario that has a scenario description:

```gherkin
Scenario: Do a thing
    In this scenario we do a thing.
    In this scenario description,
    it can be useful to do some explaining

    Given a thing
    When I do something
    Then something is done
```

In the above case, `pytest-bdd` treats each line of the description as a new scenario and creates three extra test cases `test_in_this_scenario_we_do_a_thing`, `test_in_this_scenario_description`, and `test_it_can_be_useful_to_do_some_explaining` and executes them without an error.

It may be easier to just allow scenario descriptions, but at the very least, I'd expect an error saying that the first line of a scenario must be a step.

rbcasperson:
Fixes #311 

In general, I tried to treat scenario descriptions exactly the same way as feature descriptions are currently treated. I'm not sure to what extent feature descriptions are used besides being an attribute on the `Feature` class, but the `Scenario` class now has a `description` attribute as well.

The biggest uncertainty I have with the code in this initial version of the PR is the testing. The current bug is a strange one which involves it creating extra scenarios that would be executed, and would pass with correctly implemented steps. Basically the same scenario runs as many times as there are lines in the scenario description.

The way I discovered the bug was by using the `skip` marker. The skip marker was only applied to the _first_, _real_ scenario, so the subsequent "rogue" scenarios were not skipped, and thus gave me error for not implemented steps. In the first commit, I created tests that implement that. Without the fix in the second commit, only the first test would be skipped, and the rogue ones would fail. Certainly give it a try!

I also added a test in the already existing `description.feature` and `test_description.py` files. Together, I think they sufficiently confirm that scenario descriptions are all good, but I'm happy to head suggestions for additional tests or test changes.

Thanks, and hopefully this can get moving quickly!


rbcasperson:
@youtux it's been a little while, but I've made the little tweak to get CI passing for this PR.

I've also followed along in #306, and perhaps that new work will help with some of the parsing issues. But I don't see why these changes would have to wait for something like that.

rbcasperson:
@youtux it's been a little while, but I've made the little tweak to get CI passing for this PR, but it looks like there's a 405 response error in travis when making a coveralls API call. Not sure how to address that.

I've also followed along in #306, and perhaps that new work will help with some of the parsing issues. But I don't see why these changes would have to wait for something like that.

rbcasperson:
@youtux pinging here again to see if I can get anyone to look at this PR. There are other things that I would also like to fix, but I want to be sure if I spend the time, it will actually be looked at. Thanks!

tests/feature/test_scenario_descriptions.py
```diff
@@ -0,0 +1,4 @@
+from pytest_bdd import scenarios
+
+# See feature file for reasons for executing (skipping) these tests
+scenarios("scenario_descriptions.feature")
```
youtux: instead of testing it this way, can you test it in the way described in https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins? In this way we can better assert if the plugin itself works correctly, instead of "hoping" that it generates the correct test items.
This is an example of what I mean: https://github.com/pytest-dev/pytest-bdd/blob/386ed90cec68f4a9e000728928bbb4854f9a56d4/tests/feature/test_no_sctrict_gherkin.py#L4

youtux: Also, I would not even have this test file, we can just squeeze it into tests/feature/test_description, as the feature we are testing is having a description in feature files.

rbcasperson: Thanks, I'll take a look and make that change.

rcasperson-jc:
@youtux it has been a little while, but I've updated the tests based on your suggestions. Hopefully you or someone can get a chance to look at this again. Thanks!

rbcasperson:
@youtux or any contributor, definitely looking for next steps on this one. Thanks!

jsa34:
Superseded by #749

PR #316, opened 2019-08-17
Fix Python 3.8 support

The-Compiler:
This is probably obsolete after #309, but for now seems to make it run fine at least for my project - let's see what Travis says!

### DIFF
--- a/pytest_bdd/steps.py
+++ b/pytest_bdd/steps.py
@@ -220,6 +220,8 @@ def recreate_function(func, module=None, name=None, add_args=[], firstlineno=Non
     ]
     if six.PY3:
         argnames.insert(1, "co_kwonlyargcount")
+        if sys.version_info.minor >= 8:
+            argnames.insert(1, "co_posonlyargcount")
 
     for arg in get_args(func):
         if arg in add_args:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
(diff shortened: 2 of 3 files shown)


youtux:
Thanks for the PR. Can you also update https://github.com/pytest-dev/pytest-bdd/blob/72a8764556cdecb96c8ce44861af8de1a5726c9c/tox.ini#L6 to `py{35,36,38}-pytestlatest,`?
Thanks

The-Compiler:
Done!

pytest_bdd/steps.py
```diff
@@ -220,6 +220,8 @@ def set_code(func, code):
     ]
     if six.PY3:
         argnames.insert(1, "co_kwonlyargcount")
+        if sys.version_info.minor >= 8:
+            argnames.insert(1, "co_posonlyargcount")
```
youtux: I'm checking the python documentation for code objects, and it seems that `co_kwonlyargcount` is there at least since python 3.5: https://docs.python.org/3.5/library/inspect.html?highlight=co_kwonlyargcount.

Does the test suite break if we don't do the `sys.version_info.minor >= 8` check?

youtux:
Hi @The-Compiler, I just released version 3.2.0 that includes the changes from #309. Unfortunately it supersedes your fix, sorry for the timing.

The-Compiler:
No worries! I love the changes in #309, it's probably the main reason I've been considering a rewrite/alternative to pytest-bdd for years (but never got around to it)! :+1:

Issue #204, opened 2016-12-21
upgrade pytest-bdd-2.16.0 to pytest-bdd-2.16.1 gives Plugin already registered: session


ghost:
When generating missing

`--generate-missing --feature`

all works good with
`Python 2.7.12, pytest-2.9.2, py-1.4.32, pluggy-0.3.1, pytest-bdd-2.16.0`

but when I do same thing with
` Python 2.7.12, pytest-2.9.2, py-1.4.32, pluggy-0.3.1,  pytest-bdd-2.16.1` or the latest 2.18.1

I get:

.7/site-packages/_pytest/vendored_packages/pluggy.py", line 350, in register
    (plugin_name, plugin, self._name2plugin))
ValueError: Plugin already registered: session

also fails with bdd-2.18.1 + pytest 3.0.5

matthussey:
I am getting this too. I downgraded to pytest-bdd 2.16.0 and it worked fine. I am using Python 2.7.13 on Windows 10.

matthussey:
I've had a little look and the commit that causes this behaviour to begin for me is: be79c1d534634baeacddb1098e97988c9df8e9fb

matthussey:
I've had a little look and put in some debug. It looks like pytest registers the module under test during setup and then it gets registered again during pytest-bdd running and this causes the error. I can't see why though.

olegpidsadnyi:
@pouledodue is it still a case with the newest bdd?

michaelnt:
I get this error with the latest version of pytest-bdd (3.0.1) and pytest (4.1.0)

> lib/python3.7/site-packages/pluggy/manager.py", line 80, in register
    % (plugin_name, plugin, self._name2plugin)
ValueError: Plugin already registered: session=<Session tests>

pfctdayelise:
I likewise get this error with pytest-bdd-3.1.0 and pytest 4.4.1. Nothing else in my virtualenv.

EveAndElse:
Are there any news? It's still happen using pytest-bdd-3.1.1 and pytest 5.0.1

jamesharris-garmin:
I think I found the issue and posted a comment to that effect.
https://github.com/pytest-dev/pytest-bdd/commit/be79c1d534634baeacddb1098e97988c9df8e9fb#r34811134

I will post a pull request shortly

jamesharris-garmin:
[...]

PR #319, opened 2019-08-23
Fix pytest --generate --fixtures flag

jamesharris-garmin:
This patch resolves the issues where pytest will try to double load the
pytest_cmd_main.

This resolves issue #204


jamesharris-garmin:
@youtux you might find this useful.

youtux:
Thanks for the fix. Can you also add these 2 lines at the end of test_generate_missing so that we cover it for the future?
```
assert not result.stderr.str()
assert result.ret == 0
```

jamesharris-garmin:
Done and done!

youtux:
Thanks for the fix!

PR #321, opened 2019-08-24
Refactor code coverage configuration

youtux:
Changes:

- `tox.ini` much more readable
- always run pytest with coverage (coverage is then combined)
- switch from Coveralls.io to Codecov, as it is seem to be preferred by the pytest-dev organization, and it seems to be more flexible than Coveralls.
- Add pytest 5.1.x to the list of pytests to build against


.travis.yml
```diff
@@ -1,21 +1,19 @@
...
-
+  - "/^\\d/"
+after_success:
+- codecov
 notifications:
   email:
-    - bubenkoff@gmail.com
-    - oleg.pidsadnyi@gmail.com
+  - bubenkoff@gmail.com
+  - oleg.pidsadnyi@gmail.com
```
olegpidsadnyi: maybe you could add your email here

youtux: No thanks, I get spammed enough :)

PR #327, opened 2019-09-12
Escape quote in generation

Mark90:
Hi!

I noticed that when generating testcode, single quotes would not be escaped properly. The fix was relatively easy. So easy that I'm wondering if there may have been a reason not to fix it, perhaps something breaks now? At any rate, all py37 testscases for all pytest versions seem to work fine.

I've changed generate.feature and test_generate.py to demonstrate the bug, and the second commit adds the fix and greenlights the tests again.

About the fix: I pondered on using `repr()`, but that would enclose strings in double-quotes, which is a much more noticeable change. Also, repr() is considered a debugging tool, so all in all it didn't feel right.

Hope I'm using the right branch to merge to, and the correct coding style. If not, let me know and I'll change it as needed.

Thanks!


pytest_bdd/scenario.py
```diff
@@ -292,6 +292,11 @@ def make_python_name(string):
     return re.sub(ALPHA_REGEX, "", string).lower()
 
 
+def make_string(string):
+    """Make a string literal."""
+    return "'{}'".format(string.replace("'", r"\'"))
```
youtux: From the [python docs about repr](https://docs.python.org/3.7/library/functions.html#repr),

> Return a string containing a printable representation of an object. For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval()

So I think that `repr()` is exactly what we should be using here. It will also determine whether to use single or double quotes, according to the string being passed.

tests/scripts/generate.feature
```diff
@@ -3,6 +3,7 @@ Feature: Code generation
     Scenario: Given and when using the same fixture should not evaluate it twice
         Given I have an empty list
         And 1 have a fixture (appends 1 to a list) in reuse syntax
+        And There's a step with a quote
```
youtux: This should be a test by itself, let's not hook in other tests.

Let's also test that it behaves correctly in case of double quotes.

pytest_bdd/generation.py
```diff
@@ -47,7 +47,11 @@ def generate_code(features, scenarios, steps):
     grouped_steps = group_steps(steps)
     template = template_lookup.get_template("test.py.mak")
     return template.render(
-        features=features, scenarios=scenarios, steps=grouped_steps, make_python_name=make_python_name
+        features=features,
```
youtux: This library now uses [black](https://github.com/psf/black) as the code formatter.
Make sure you run `black` against the committed files (I use `pre-commit` hooks to automatically detect if that was not done, more info [here](https://github.com/psf/black#version-control-integration))

pytest_bdd/scenario.py
Mark90: Yes, it's the better solution. However it will create double-quoted strings, so for consistency I think [scenario.feature.rel_filename](https://github.com/pytest-dev/pytest-bdd/blob/b561f590b553eea49664a48585afdbcc3d32dcc1/pytest_bdd/templates/test.py.mak#L15) should also be double-quoted. (which is something I let black fix after running generate anyway :-) )

What do you think?

youtux: Yes, definitively.
`repr()` will try to make single-quote strings, and fallback to double-quote strings in case there is already a single quote it the source.
We should not care about `black`ening the resulting files, as they will not be committed in the repository.

Mark90: Ah, I phrased that poorly. What I meant was that when I use `pytest-bdd generate` in my daily work, I usually run black right after.
Anyway, I'll remove the make_string and just use repr inline in the template.

tests/scripts/generate.feature
Mark90: Ok, where do I put it? As another scenario in generate.feature, or a new feature to check the quote escaping of generate (`generate_escape_quotes.feature` or something)?

pytest_bdd/generation.py
Mark90: Done. It also reformatted another file which I've committed separately.

tests/scripts/generate.feature
Mark90: Added a separate generate_with_quotes.feature and added a testcase in test_generate.py

pytest_bdd/scenario.py
Mark90: `make_string` removed in favor of repr. Then I added make_python_docstring to handle `"""triple"""` quoted strings in docstrings. (overkill? maybe :)) See generate_with_quotes.feature for examples.

Mark90:
Welp, some error in the 3.7 test-suite that I don't understand, related to formatting. And the Python 2.7 tests fails on test_generate.py because it expects unicode literals. I don't recall it doing that a month ago, but I'm probably wrong.

When you're ok with the code I'll look into those test issues and create a new branch/PR with 1 proper commit. Too many commits in this one now, and I set the wrong author on them as well.

tests/scripts/test_generate.py
```diff
@@ -60,3 +60,69 @@ def my_list_should_be_1():
...
+
+    @then('The generated string should be written')
+    def the_generated_string_should_be_written():
+        """The generated string should be written."""
+        raise NotImplementedError
+
+    '''[
+            1:
+        ].replace(
+            u"'", u"'"
```
youtux: what is this replace doing?

tests/scripts/test_generate.py
```diff
@@ -60,3 +60,69 @@ def my_list_should_be_1():
...
+        raise NotImplementedError
+
+
+    @then('The generated string should be written')
+    def the_generated_string_should_be_written():
+        """The generated string should be written."""
+        raise NotImplementedError
+
+    '''[
+            1:
```
youtux: in order to skip the first new line, you can define the expected string like this:
```python
    assert out == textwrap.dedent(
        '''\
    # coding=utf-8
```

tests/scripts/test_generate.py
```diff
@@ -60,3 +60,69 @@ def my_list_should_be_1():
             u"'", u"'"
         )
     )
+
+
+def test_generate_with_quotes(monkeypatch, capsys):
+    """Test that generated code escapes quote characters by a given feature."""
+    monkeypatch.setattr(sys, "argv", ["", "generate", os.path.join(PATH, "generate_with_quotes.feature")])
```
youtux: let's not monkey patch.
You can probably use `testdir.run()` instead.

tests/scripts/generate_with_quotes.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Handling quotes in code generation
```
youtux: I would instead define this whole feature file in a docstring in test_generate_with_quotes and write it to a file in the testdir programmatically, just like we do in https://github.com/pytest-dev/pytest-bdd/blob/f4dce04053036f10350ce682396c34c6d8cc4dec/tests/feature/test_no_sctrict_gherkin.py#L4

youtux:
Also don't worry about the number of commits, I can squash them anyway when approving.

Mark90:
Thanks for your comments, using pytest.testdir now.
And resolved the formatting issue by updating my black version from 18.9 to 19.3 and rerunning it

Just the python 2.7 test failing,looking into that

Mark90:
Had a good bit of fun fixing the Py2 / unicode literal. Finally decided to move away from `repr()` as there doesn't seem to be a way to use it in Python 2, at least not without introducing some form of regression (either the string literals would become unicode literals, or unicode characters wouldn't be rendered anymore because of repr) 

Also, with the code as it is now, quotes are escaped the same in Python2/Python3, which I found a better tradeoff. But still, I'm not sure if maintaining Py2 compatibility is such a priority, given its nearing EOL date. What do you think?

I also included a fix for Py2 when running generate on a feature file with unicode characters, which would raise UnicodeEncodeError when piping the output.

youtux:
Thanks for the pull request!

PR #332, opened 2019-10-10
Fix for Python 4

hugovk:
We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule) and work begins on Python 3.9+1.

There's some code which uses `six.PY3`:

```python
if six.PY3:
    long = int
```

Where:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this skip the line! Instead, use `not six.PY2`.

Found using https://github.com/asottile/flake8-2020:
```console
$ pip install -U flake8-2020
$ flake8 --select YTT
```

### DIFF
--- a/pytest_bdd/cucumber_json.py
+++ b/pytest_bdd/cucumber_json.py
@@ -7,12 +7,7 @@ import os
 import sys
 import time
 
-import six
-
 from .feature import force_unicode
-
-if not six.PY2:
-    long = int
 
 
 def add_options(parser):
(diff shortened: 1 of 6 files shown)


pytest_bdd/cucumber_json.py
```diff
@@ -11,7 +11,7 @@
 
 from .feature import force_unicode
 
-if six.PY3:
+if not six.PY2:
```
youtux: Let's instead remove usages of `long` completely and replace them with `int`.

youtux:
The CI build is failing because `attrs` 19.2.0 was recently released, and it has some changes that are incompatible with pytest between 4.0 and 4.2.
We should probably drop compatibility with pytest < 4.3 then.

hugovk:
Thanks, updated!

youtux:
Can you also add an entry to the changelog? Thanks

hugovk:
You're welcome!

Issue #324, opened 2019-09-02
--cucumberjson-expanded does not work for pytest.mark.parametrize


fundakol:
Step name does not contain argument value in scenario outline in Cucumber JSON report when I use @pytest.mark.parametrize decorator.

Version:
python==3.7.3
pytest==5.1.2
pytest-bdd==3.2.1

```gherkin
# -- FILE: scenario_outline.feature
Scenario Outline: Outlined given, when, thens
    Given there are <start> cucumbers
    When I eat <eat> cucumbers
    Then I should have <left> cucumbers
```

```python
# -- FILE: scenario_outline_test.py
import pytest
from pytest_bdd import given, when, then, scenario


@pytest.mark.parametrize(['start', 'eat', 'left'],
                         [(15, 7, 8),
                          (15, 5., 10)])
@scenario(
    'scenario_outline.feature',
    'Outlined given, when, thens',
    example_converters=dict(start=int, eat=int, left=int)
)
def test_outlined(start, eat, left):
    pass


@given('there are <start> cucumbers')
def start_cucumbers(start):
    return dict(start=start)


@when('I eat <eat> cucumbers')
def eat_cucumbers(start_cucumbers, eat):
    start_cucumbers['eat'] = eat


@then('I should have <left> cucumbers')
def should_have_left_cucumbers(start_cucumbers, start, eat, left):
    assert start - eat == left
    assert start_cucumbers['start'] == start
    assert start_cucumbers['eat'] == eat
```    

I run the test with following options:
```shell
python -m pytest scenario_outline_test.py -s -vvv --cucumberjson=cucumber.json --cucumberjson-expanded
```

Here is my output from cucumber.json:
```json
"steps": [
          {
            "keyword": "Given",
            "name": "there are <start> cucumbers",
            "line": 2,
            "match": {
              "location": ""
            },
            "result": {
              "status": "passed",
              "duration": 996589
            }
          },
```
It works fine when I use Example in the feature file.

marcbrossaissogeti:
Hello,
i have an issue and i think it is the same as you.
I try to generate a cucumberjson file by using :
` py.test test_feature_sample.py --cucumberjson=cucumberjson.json --cucumberjson-expanded`

I get the following error:
> pytest_bdd\cucumber_json.py", line 111, in _format_name
>   name = name.replace('<{}>'.format(param), value)
>TypeError: replace() argument 2 must be str, not 

[...]

PR #335, opened 2019-10-28
cast "values" as a string because it could be another type if using e…

marcbrossaissogeti:
#324 --cucumberjson-expanded does not work for pytest.mark.parametrize

To make it clearer:
The method "_format_name" in cucumber_json.py is using another method which expect all parameters to be strings. If you use example converters, the argument could be an other type.

I have fixed the behavior by only casting the parameter : 
```python
    name = name.replace("<{}>".format(param), str(value))
```


marcbrossaissogeti:
up ? @youtux it should be easy to integrate

youtux:
Hi @marcbrossaissogeti, thank you for the fix. Can you add a test for it?

ishmelev23:
The problem still exists. Why it hasn't been merged, yet?

youtux:
@ishmelev23 I missed the updates to this PR. It's merged now.
@marcbrossaissogeti thanks for the PR.

Issue #333, opened 2019-10-14
Escape pipe character ("|") in scenario outline


Mark90:
Hi, is there any way to escape the pipe character in scenario outlines? The sample code below shows the "problem":

**sample.feature**
```feature
Feature: Using special characters

    Scenario Outline: Testing special characters
        Given We have strings <string1> and <string2>
        When We compare them
        Then They should be equal

        Examples:
        | string1    | string2    |
        | bork       | bork       |
        | bork\|bork | bork\|bork |
```

**test_sample.py**
```python
"""Using special characters feature tests."""

from pytest_bdd import given, scenario, then, when


@scenario("sample.feature", "Testing special characters")
def test_testing_special_characters():
    """Testing special characters."""


@given("registry")
def registry():
    return dict()


@given("We have strings <string1> and <string2>")
def we_have_strings_string1_and_string2(string1, string2, registry):
    """We have strings <string1> and <string2>."""
    registry["string1"] = string1
    registry["string2"] = string2


@when("We compare them")
def we_compare_them(registry):
    """We compare them."""


@then("They should be equal")
def they_should_be_equal(registry):
    """They should be equal."""
    assert registry["string1"] == registry["string2"]
```

When running the test;

```
(pytest-bdd-37) Ξ python/pytestbddpipechar → pytest
================================================================== test session starts ===================================================================
platform darwin -- Python 3.7.2, pytest-5.2.1, py-1.8.0, pluggy-0.13.0
rootdir: /Users/mark/dev/python/pytestbddpipechar
plugins: bdd-3.2.1
collected 0 items / 1 errors

========================================================================= ERRORS =========================================================================
____________________________________________________________ ERROR collecting test_sample.py _____________________________________________________________
test_sample.py::test_testing_special_characters: in "parametrize" the number of names (2):
  ['string1', 'string2']
must be equal to the number of values (4):
  ['bork\\', 'bork', 'bork\\', 'bork']
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
==================================================================== 1 error in 0.08s ====================================================================
(pytest-bdd-37)
```

[...]

PR #337, opened 2019-11-01
Implement parsing escaped pipe characters in outline parameters

Mark90:
Proposed implementation for https://github.com/pytest-dev/pytest-bdd/issues/333

Let me know what you think!

### DIFF
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -59,7 +59,7 @@ STEP_PREFIXES = [
 
 STEP_PARAM_RE = re.compile(r"\<(.+?)\>")
 COMMENT_RE = re.compile(r"(^|(?<=\s))#")
-EXAMPLE_PARAM_RE = re.compile(r"(?<!\\)\|")
+SPLIT_LINE_RE = re.compile(r"(?<!\\)\|")
 
 
 def get_step_type(line):
@@ -142,16 +142,14 @@ def get_tags(line):
     return set((tag.lstrip("@") for tag in line.strip().split(" @") if len(tag) > 1))
 
 
-def get_params(line):
-    """Get parameters out of the given Examples line.
+def split_line(line):
+    """Split the given Examples line.
 
     :param str|unicode line: Feature file Examples line.
 
-    :return: List of parameters.
+    :return: List of strings.
     """
-    if not line:
-        return []
-    return [cell.replace("\\|", "|").strip() for cell in EXAMPLE_PARAM_RE.split(line[1:-1])]
+    return [cell.replace("\\|", "|").strip() for cell in SPLIT_LINE_RE.split(line[1:-1])]
 
 
 def get_features(paths, **kwargs):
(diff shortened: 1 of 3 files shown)


Mark90:
I tried the 3.8 tox build locally before I made this PR, it reported the same black formatting error as travis just did. It's in `pytest_bdd/scripts.py` which hasn't been changed.
I recall having a similar issue for that file, when my local black changed `print ()` to `print()`, while CI wanted the former. Now it wants the latter. 
Not sure how to proceed :)

Mark90:
I opted to just apply the black formatting to the offending file. Seems that the CI accepts the formatting now. However, both 3.5 and 3.6 testsuites report an error in `tests/feature/test_gherkin_terminal_reporter.py`


> E           AssertionError: assert '============================== 1 passed in 0.03s ===============================' == '============================== 1 passed in 0.02s ==============================='


This seems kind of arbitrary.

The-Compiler:
It fails here:

https://github.com/pytest-dev/pytest-bdd/blob/e87b7563063c4c68e6daf1a9ee57de7f8e2a3995/tests/feature/test_gherkin_terminal_reporter.py#L175-L181

That code seems to account for that, but with pytest 5.1.0, the output format [changed](https://docs.pytest.org/en/latest/changelog.html#id76), breaking that normalizing...

The-Compiler:
#340 should fix the test.

Mark90:
@The-Compiler Thanks for the quick fix!

Will look into the change reported by codecov.

pytest_bdd/feature.py
```diff
@@ -141,6 +142,18 @@ def get_tags(line):
     return set((tag.lstrip("@") for tag in line.strip().split(" @") if len(tag) > 1))
 
 
+def get_params(line):
```
youtux: I think that "split_line" would be a better name for this, since we are not getting parameters from a line, but just splitting it with handling of escape sequences.

pytest_bdd/feature.py
```diff
@@ -59,6 +59,7 @@
 
 STEP_PARAM_RE = re.compile(r"\<(.+?)\>")
 COMMENT_RE = re.compile(r"(^|(?<=\s))#")
+EXAMPLE_PARAM_RE = re.compile(r"(?<!\\)\|")
```
youtux: I would call this `SPLIT_LINE_RE`

tests/feature/test_outline.py
```diff
@@ -166,3 +166,71 @@ def test_outlined_feature(request):
         ["fruits"],
         [[u"oranges"], [u"apples"]],
     )
+
+
+def test_outline_with_escaped_pipes(testdir):
```
youtux: The "when" and "then" steps are doing nothing, but we could improve the test by having "string1" being the string one under test, and string2 as the base64 representation of string1, and the "then" step can be something like "then <string2> should be the base64 encoding of <string1>.

The feature file would look like this (note the use of `r""" """` string, it allows to use less backslashes):
```
r"""\
    Feature: Outline With Special characters
        Scenario Outline: Outline with escaped pipe character
            Given we have strings <string1> and <string2>
            Then <string2> should be the base64 encoding of <string1>
            Examples:
            | string1        | string2          |
            | bork           | Ym9yaw==         |
            | \|bork         | fGJvcms=         |
            | bork \|        | Ym9yayB8         |
            | bork\|\|bork   | Ym9ya3x8Ym9yaw== |
            | \|             | fA==             |
"""
```

Let's also have these examples, where we check some edge cases with the escaping:
```
| bork \\\| | Ym9yayBcfA== |
| bork \\   | Ym9yayBc     |
```

Mark90: Great idea! Just implemented and pushed it. The edge-case with 3 backslashes had a different base64 encoding actually, because the 3rd backslash is consumed by `split_line()`.

pytest_bdd/feature.py
Mark90: That makes a lot more sense, yes. Also removed the `if not line` check I had put in there, which I think was unnecessary.

pytest_bdd/feature.py
Mark90: Done

youtux:
Thank you for the PR, @Mark90!

PR #340, opened 2019-11-20
Fix gherkin terminal reporter test flakiness

The-Compiler:
The test compares the output of two different reporters and tries to account for
the different runtime being printed. However, the output format changed with
pytest v5.1.0:

https://docs.pytest.org/en/latest/changelog.html#id76

This changes the regex to be a bit more permissive.


youtux:
Thank you for the fix!

PR #345, opened 2020-01-10
Override language

napolas:
When we override the language by redefining the "types" constants, the terminal reporting as well as the cucumber.json output does not follow as expected.
Although this is not an I18N proper solution, it allows one to override the types constants and even so the reporting tools work in consistent fashion.


youtux:
Superseded by https://github.com/pytest-dev/pytest-bdd/pull/725

PR #347, opened 2020-02-18
Creates path to json file if it does not exist

shawnemhe:
Minor change to prevent raising an error when attempting to create a cucumber json in a directory that has not been created yet. This breaks form with the standard pytest --junitxml functionality which would create the path in this situation.


youtux:
Hi @shawnemhe, thank you for this fix. Could you write a test for it too?

shawnemhe:
Hi @youtux, it is testing now by parametrizing the testdir fixture so that of the cucumber tests run with and without a missing subdirectory.

Issue #223, opened 2017-09-18
Async Step Definitions


jeduden:
is there away to use async step definitions with pytest-bdd ?

For a step definition like:
`
@when('i send cucumbers ')
async def i_send_cucumbers(loop):
      pass
`
I see the warning:
`
pytest_bdd/scenario.py:137: RuntimeWarning: coroutine 'i_send_cucumbers' was never awaited
`

olegpidsadnyi:
what about async then? does it even make sense?

olegpidsadnyi:
will this work for you?


async def send_cucumbers():
    pass


@when('i send cucumbers ')
def i_send_cucumbers(loop):
    loop.run_until_complete(send_cucumbers)

jeduden:
This works. However, in my case all step definitions are async (and can call multiple times into async functions). so this would mean a lot of calls to run_until_complete.
As a work-around I have defined a decorator that schedules the function on the event loop.

A generic solution would be to have the ability to easily wrap any step definition with a function. (similar to the before and after hooks).

And then it would be great if support of async functions would be provided out of the box. 

Especially, after reading the discussion about implicit and explicit eventloops in the thread here: https://groups.google.com/forum/#!msg/python-tulip/yF9C-rFpiKk/tk5oA3GLHAAJ seems to be tending towards implicit loops for python code that runs on 3.5.3+ and/or 3.6+. 

So code that explicitly calls run_until_complete to run async functions will look more and more awkward.

olegpidsadnyi:
@jeduden this means pytest-bdd has to depend on yet another library asyncio? I don't really see how a gherkin scenario can by async. If the whole point that it has not to break i think you need to extend your testing suite with some kind of support for async functions (decorator is a good idea).

I think semantically you can't really represent async process in step-by-step imperative Gherkin. Therefore it should be explicit. If you have to initiate few async messages - they should stand behind one When step that describes this exercise in a form that humans understand. Gherkin is not a programming language, it is the way to describe steps to humans which can't do async computation in their brains.

What do you think, @bubenkoff ?

bubenkoff:
well, the async question comes down to the pytest, not specifically to pytest-bdd.
[...]

PR #349, opened 2020-03-12
Add support for async steps & scenarios

amadeuszhercog-silvair:
Relates to https://github.com/pytest-dev/pytest-bdd/issues/223

This pull request add support for asynchronous steps and scenario functions. As i wrote in `README.rst` user will be able to use async steps and scenarios out of the box like this:
```
@pytest.mark.asyncio
@scenario('test.feature', 'Launching scenario function')
def test_launching_scenario_function():
    pass

@given("i have async step")
async def async_given():
    pass

@when("i do async step")
async def async_when():
    pass

@then("i should have async step")
async def async_then():
    pass
```

### Explanation about tests

I've put all tests that use `asyncio` into `tests/asyncio`. Their structure is following:
* `test_async_given_returns_value.py` - tests checking that async `given` is a fixture or it can shadow other fixture,
* `test_async_scenario_function.py` - tests for scenario function,
* `test_async_steps.py` - tests for async steps,
* `test_launching_app_in_background.py` - (bigger than above ones) test that will launch dummy application based on asyncio in `event_loop.create_task` and will interact with it by http protocol. It checks that app is still running between steps and it can respond.
* `dummy_app.py` - dummy app from previous point + flask app with which is "gateway" between dummy app and tests.

Data flow of tests based on dummy app looks like this:
```
test function -> flask server -> dummy app -> flask server -> test function
```

**That PR will require dropping Python 2.7 support before.** But since Python 2.7 is no longer officially supported anyway I don't think it would be a problem.


yanlend:
I don't think the test with the app is worth to add all those testing requirements.

Otherwise, I like the PR. I think adding a native support for asyncio is useful!

mikeoconnor0308:
Are there any plans to look at getting this merged?

youtux:
Hi @amadeuszhercog-silvair, please check my [comment](https://github.com/pytest-dev/pytest-bdd/pull/426#issuecomment-869122229) about trying to use a different approach for this issue.

It would be better to have async code compatibility without being tied to a specific pytest plugin.

PR #356, opened 2020-04-20
Possible to relax Gherkin validation for simple steps generator.

v-buriak:


pytest_bdd/scripts.py
```diff
@@ -47,20 +47,38 @@ def check_existense(file_name):
...
+
+
 def main():
     """Main entry point."""
     parser = argparse.ArgumentParser(prog="pytest-bdd")
     subparsers = parser.add_subparsers(help="sub-command help", dest="command")
     subparsers.required = True
     parser_generate = subparsers.add_parser("generate", help="generate help")
+    parser_generate.add_argument(
+        "--strict-gherkin",
```
olegpidsadnyi, line 75: Why not just --strict-gherkin without the True|False, should default=False and store_true do the job?
Ah.. It is strict by default...

youtux:
I think that the steps generator can just always work in non-strict mode.

olegpidsadnyi:
Yes, passing non strict always to the feature parsing is also an option

v-buriak:
> Yes, passing non strict always to the feature parsing is also an option

But if we will change default behaviour of generator then we will confuse beginners. In one case for generator default behaviour accept any order and later they will encounter an error with pytes run.

v-buriak:
What to do with this?
Some checks haven’t completed yet
2 expected and 1 successful checks

"codecov/patch Expected — Waiting for status to be reported"
"codecov/project Expected — Waiting for status to be reported"

olegpidsadnyi:
@v-buriak let's just pass strict=False and remove the option. It looks like nobody really cares about it being strict or not.

youtux:
RE codecov, it seems that they released a buggy version: https://github.com/codecov/codecov-python/issues/249

v-buriak:
> @v-buriak let's just pass strict=False and remove the option. It looks like nobody really cares about it being strict or not.

ok

youtux:
@v-buriak can you please add an entry to the changelog, under the "unreleased" section?

v-buriak:
> @v-buriak can you please add an entry to the changelog, under the "unreleased" section?

PR is merged, should i open new one?

olegpidsadnyi:
@v-buriak no worries. I'll add it and do a release when i have a miniute

olegpidsadnyi:
@v-buriak can you check the release 3.3.0 if it works for you?

v-buriak:
> @v-buriak can you check the release 3.3.0 if it works for you?

Yes, I will check.

v-buriak:
> > @v-buriak can you check the release 3.3.0 if it works for you?
> 
> Yes, I will check.

It works for me, thanks!

PR #361, opened 2020-05-11
Generate missing with parsers

david-vankampen:
The `--generate-missing` argument to pytest does not recognize given/when/then clauses that are implemented using a parser from the parsers module. However, the generation code was discovering them.  It just wasn't returning it as found due to a duplicate name prefix being applied.  

This PR removes that prefix in those cases, which allows `--generate-missing` to recognize parser clauses as NOT missing.


bissellmos:
@youtux checking in on this one - anything else I can do?

david-vankampen:
@youtux checking in on this one - anything else I can do?

AntoineDao:
@david-vankampen this should now be resolved through #415.

PR #180, opened 2016-04-19
Add support for data tables

PatrickMassot:
Fixes #150 and helps mitigating the effects of #157 

The goal is to allow data tables attached to steps as described in https://cucumber.io/docs/reference#data-tables. This has nothing to do with example tables, it does not trigger multiple run of the scenario. An artificial example would be:

```
Feature: Datatables

  Scenario Outline: I use datatables
    Given the following users exist:
      | name   | email              | twitter         |
      | Aslak  | aslak@cucumber.io  | @aslak_hellesoy |
      | Julien | julien@cucumber.io | @jbpros         |
      | Matt   | matt@cucumber.io   | @mattwynne      |
    Then I should see the following names:
      | name   |
      | Aslak  |
      | Julien |
      | Matt   |
```

For each step which has a data table, the implementation can use a datatable argument which is a list of two lists: the list of headers and the list of rows. 

```
@given('the following users exist:')
def the_following_users_exist(datatable):
    """the following users exist:."""
    return datatable[1]

@then('I should see the following names:')
def i_should_see(datatable, the_following_users_exist):
    names = [row[0] for row in the_following_users_exist]
    assert names == sum(datatable[1], [])

```
## Implementation

Commits should be reviewed one by one, otherwise things will be cluttered. Note that all commits are fully toxed.
- 9f744fd performs clean up that have nothing to do with the issues.  It should probably be merged in any case.
- 8461566 is preparatory. Since a data table very much looks like an example table I factored out a generic class `Table`.  Not much happens here, only renaming and trivial subclassing (I chose to keep the class `Examples` for clarity and debugging purposes, although it inherits absolutely everything from `Table`).
- 0b14878 implements the feature parsing part. Datatables are sort of multi-line steps so they are parsed at the same point in the file. I also introduced two helper functions `is_table_line` and `parse_table_line`.
- 454a01a is where I could easily be totally wrong. Fixture injection is deep into python dark magic and totally out of my comfort zone. Still I tried to figure out how to pass the data tables to steps implementations
- 67fffa1 adds some minimal test for this new feature (in case of merge we should probably add more)
- 7e7e9a7 is because I missed one  `git commit --amend` before pushing...

---

This change is [Reviewable](https://reviewable.io/reviews/pytest-dev/pytest-bdd/180)

Issue #157, opened 2015-10-12
pytest_bdd.exceptions.GivenAlreadyUsed with step arguments


The-Compiler:
From the readme:

> Often it's possible to reuse steps giving them a parameter(s). This allows to have single implementation and multiple use, so less code. Also opens the possibility to use same step twice in single scenario and with different arguments!

But this doesn't seem to be the case for `@given`. My usecase is this feature file:

``` cucumber
Feature: Going back and forward.
    Testing the :back/:forward commands.

    Scenario: Going back
        Given I open backforward/1.html
        And I open backforward/2.html
        When I run :back
        Then backforward/1.html should be loaded
```

and this python file:

``` python
import pytest_bdd as bdd

bdd.scenarios('.')


@bdd.given(bdd.parsers.parse("I set {sect} -> {opt} to {value}"))
def set_setting(quteproc, sect, opt, value):
    quteproc.set_setting(sect, opt, value)


@bdd.given(bdd.parsers.parse("I open {path}"))
def open_path(quteproc, path):
    quteproc.open_path(path)


@bdd.when(bdd.parsers.parse("I run {command}"))
def run_command(quteproc, command):
    quteproc.send_cmd(command)


@bdd.then(bdd.parsers.parse("{} should be loaded"))
def url_should_be_loaded(httpbin, path):
    requests = httpbin.get_requests()
    assert requests[-1] == [('GET', url)]
```

So opening those two pages is clearly "setup", i.e. I think they both belong to `Given ...`.

But I get:

```
pytest_bdd.exceptions.GivenAlreadyUsed: Fixture "open_path" that implements this "I open backforward/2.html" given step has been already used.
```

olegpidsadnyi:
The idea behind "given i have an A" is that there's some fixture that implements A and stores it's instance for later steps. Following steps could use it as "When I go to the A's edit page" which means you would get fixture as a dependency.
This way - there's only one instance of A. Basically saying "given I have an A" multiple times doesn't make sense. In my opinion you can't take incremental side effect action in the given section.
In your case you say "I open" and this is an "exercise" by definition and exercises belong to the when section.

The-Compiler:
I see, that makes sense - thanks for your explanation!

spinus:
If I occasionally need to use same given multiple times  I use container. Example:
```
@pytest.fixture
def basket():
  return {}

@given("There is {var}")
def there_is_var(basket, var):
  basket[var] = "something"
  return basket

[...]

PR #366, opened 2020-06-05
Implement datatables

### LINKED ISSUE #150: Datatables
Hello,
support for datatables would be a big help. 
See following example from the cucumber reference.
https://cucumber.io/docs/reference#data-tables

Thanks a lot.

gnikonorov:
Implement datatables for `Given` clauses. The work for this was mainly done in https://github.com/pytest-dev/pytest-bdd/pull/180. I am recreating it since I am taking it over to help get this merged in. The changes are identical between the two. All I did was merge in the latest changes to the PR/resolved merge conflicts and added more testing.

@youtux @olegpidsadnyi can we get this reviewed/merged in please? It looks like a lot of the community wants this feature ( myself included ) based off of the comments in the original PR.

Copied from origin PR: Fixes #150 and helps mitigating the effects of #157

### DIFF
--- a/pytest_bdd/feature.py
+++ b/pytest_bdd/feature.py
@@ -1,9 +1,7 @@
 """Feature.
 
 The way of describing the behavior is based on Gherkin language, but a very
-limited version. It doesn't support any parameter tables.
-If the parametrization is needed to generate more test cases it can be done
-on the fixture level of the pytest.
+limited version.
 The <variable> syntax can be used here to make a connection between steps and
 it will also validate the parameters mentioned in the steps with ones
 provided in the pytest parametrization table.
@@ -19,8 +17,6 @@ Syntax example:
         And the article should be published  # Note: will query the database
 
 :note: The "#" symbol is used for comments.
-:note: There're no multiline steps, the description of the step must fit in
-one line.
 """
 
 from collections import OrderedDict
@@ -143,9 +139,9 @@ def get_tags(line):
 
 
 def split_line(line):
-    """Split the given Examples line.
+    """Split the given table line.
 
-    :param str|unicode line: Feature file Examples line.
+    :param str|unicode line: Feature file table line.
 
     :return: List of strings.
     """
@@ -174,15 +170,35 @@ def get_features(paths, **kwargs):
     return features
 
 
-class Examples(object):
+def is_table_line(line):
+    """Detect table line.
 
-    """Example table."""
+    :param `string` line: the line to inspect
+
+    :return: `bool`.
+    """
+    return line.startswith("|") and line.endswith("|")
+
+
+def parse_table_line(line):
+    """Parse cells contained in a table line.
+
+    :param `string` line: the line to parse (stripped)
+
+    :return: `string list`.
+    """
+    return [l for l in split_line(line)]
+
+
+class Table(object):
+
+    """Example or data table."""
 
     def __init__(self):
-        """Initialize examples instance."""
-        self.example_params = []
-        self.examples = []
-        self.vertical_examples = []
+        """Initialize table instance."""
+        self.params = []
+        self.entries = []
+        self.vertical_entries = []
         self.line_number = None
         self.name = None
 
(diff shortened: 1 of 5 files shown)


olegpidsadnyi:
@gnikonorov me and @youtux have been discussing these PR of yours and #180 
We are thinking of implementing this support for both "docstring" and "datatables" in a different way. It seems it has to be supported on the feature level as well for the cucumber.json reporting reasons.
So we'll pick this problem up and try to solve it combined with the """dosctring""" problem that @youtux  was already working on.

gnikonorov:
Thanks @olegpidsadnyi. Out of curiosity is there an ETA for this? Not having this functionality is preventing me from using `pytest-bdd` which is a shame since I would prefer to use something from `pytest-dev` over a different solution

olegpidsadnyi:
@gnikonorov this is the highest priority now

gnikonorov:
Thank you @olegpidsadnyi! 
This is unrelated but is there a contributing guide? I was looking but couldn’t find a CONTRIBUTING.rst/md

olegpidsadnyi:
@gnikonorov the project is currently is the state that is a bit hard to contribute. But we want to spend some time to clean it up and get back to maintenance. Stay tuned

gnikonorov:
Thanks @olegpidsadnyi, looking forward to contributing!

anniepapa:
@gnikonorov The status shows PR is closed but apparently this datatable feature is not merged into pytest-bdd master, which is confused. 
I found the test_datatable.feature from your repo: gnikonorov:issue_150.
When can I expect this be a part of pytest-bdd? 
This is a very useful feature... Especially for doing testing for database interaction.

olegpidsadnyi:
@anniepapa we are currently researching the possibility to switch to the official gherkin parser, that could change everything. Most of the issues in pytest-bdd are related to the differences between ours and the cucumber parser. Last weeks I was converting tests to pytester way to make it more maintainable. Stay tuned.

anniepapa:
@olegpidsadnyi Thanks for the reponse. Looking forward the coming update. 
Will this switch also provide flexibility to reuse Given step in the same scenario? Which can be then used in datawarehousing enviornment for testing.

olegpidsadnyi:
@anniepapa i'm not sure about using given multiple times. This is not related to the parser limitations, but more to the pytest paradigm.
Pytest makes it easy to refer to something you set up in other steps.
Given foo value is 5
When I print foo
Then I see 5

So you can operate a target_fixture with the name "foo".
While multiple given lines about the same type of thing.
like:
given I have inserted a record of Model with the name 1
and I have inserted a record of Model with the name 2
and I have inserted a record of Model with the name 3

How do you refer to the 2nd one? By index? So maybe these datatables is something that represents your multiple given steps.
I'm using flavors of fixtures like user, second_user, third_user

Normally 2-3 flavors is enough to test all your list views. And you still can refer to them by their symbolic names

anniepapa:
@olegpidsadnyi  : ) I have saw quite some debates regarding this. And I understand that the reuse of GIVEN is not suggested because this gives tolerance for dev/tester to write unnecessary duplicated steps. 
That's also why I mentioned this can be useful in some specific env, e.g: DWH environment. 
For example, a business requirements has 15 tables involved with various common table expression and many columns used as flag/indication for business logics. Different tables are joined by keys. In this case, reusable GIVEN + Data table is very handy to create scenarios & test cases with high readability than making step more abstract or prepare test data from fixture/background.
As I know this is supported in SpecFlow and behave. In my current organization, we like Pytest, especially the advantage to combine fixture/confest with pytest-bdd but testing for database interaction has some limitations.

olegpidsadnyi:
@anniepapa that is what i'm struggling to understand as well. You are talking about tables related by the keys. does it mean the very same test can't be run twice unless the storage is completely flushed?
BDD is about high-level specs. It is functional, the behavioural is in the name. So you assert the result after taking the exercise. That's why Given-When-Then. Most of the time the exercise is applied to a single entity. So don't you really need 1 fixture for this?
Especially setting up your tests with relations. Is it not something that you'd like to automate?

We are using pytest in combination with pytest-factoryboy. Most of our data is generated by Faker accordingly to the types. In order to make a valid test you need to make sure to setup a crucial attribute of your entity you are applying your exercise to. This is what spec means. I doubt that dumps of tables are readable or make any sense. When you see dumps you need to compute it in your head. This is exactly opposite to storytelling. This is exactly opposite to the .feature format.

Do you have a small example of the feature that you mention above?

anniepapa:
Feature: Test mapping of stored procedure abc
  In order to get new price and rate
  Stored procedure spabc_TtoL should map keys or values correctly

  Scenario: example
    Given Table: <abc> with test data:
| [key]     | sum_eur                               | quantity           |name      |
| 132521 | 0.0100000000000                 | 1                     |A220_W |
| 132522 | 0.0200000000000                 | 1                     |A230_X   |

    And Table: <def> with test data:
|name      | percent_variance |
|A220_W   | 1                |
|A221_X    | 3                |

And table <...> with test data:

When execute SP: <spabc_LtoODS>

Then Table: <rst_MAKE_ODS> has exact match
| key       | currency | adjusted_price | adjusted_price_new_fx |
| 132521| XYZ         | 0.0            | 0.01                  |
| 132522 | UVW      | 0.0            | 0.04                  |

And Table: <> has exact match ...

@olegpidsadnyi what's your suggestion to use pytest-bdd for this scenario with better readability?

olegpidsadnyi:
@anniepapa I don't even think BDD is good for this kind of testing. This is a full blackbox without a variety of the exercise. It is just one - when kaboom. And the table of truth to it.
Gherkin was designed for less technical people to be able to tell the user story.
So the fact that this is .feature syntax and not .py syntax doesn't help much.

I'd start with that you don't have to observe 2 at the same time. Unless you are making an integration test to make sure your spabc_LtoODS iterates over the table.

So maybe you need something like
@pytest.fixture
def ods(ods__sum_eur, ods__quantity, ods__name):
    return ODS(ods__sum_eur, ods__quantity, ods__name)

same for the second_ods if you need it.

Then you have steps like
given(parsers.parse("ODS sum is {value}"), target_fixture="ods__sum_eur")

and then finally the feature file in the simple case:

Feature: Less blackbox
    Scenario: Apply computations
          Given ODS sum is 30 EUR
          And ODS percentage is 1
          And there's ODS
          When I compute all the things
          Then Adjusted ODS price should be 0.01. # Looks into another table by using ods.id (ods fixture)

In more complex scenario if you want to see 2 records were processed you can mention "second ODS". This way you don't depend on 132521 value, it is your autoincrement

olegpidsadnyi:
@anniepapa The beauty of Scenarios is that you give them names for a situation.
I think instead of a table of truth you need to identify a typical test case and make a scenario for it. And your test should operate 1 record at the time.
In addition you can have a test with 2 records to make sure it is processed in the loop. And this way you have your coverage. Once you have a regression you would add a scenario.

Try to come up with scenario names for each record of your table instead. That instance should represent a scenario setup.

anniepapa:
@olegpidsadnyi 
Thank you for your suggestion and proposal.
Your idea is to design scenario vertically which will work, also with outline style. But this kind of scenario is often includes elaborately designed test case(s) on level of each row horizontally. So 1 Feature file has multiple scenarios, each scenario can have multiple test cases. The way of pytest-bdd is hard to support 1 scenario has multiple test cases, but make each scenario more unique. I can see this is two different way of thinking. But I don't agree "BDD is not good for this kind of testing"... Anyway I will stay tuned and try pytes-bdd if new updates. : )

olegpidsadnyi:
@anniepapa you know what? This restrictions make sense only when you are using fixture or target fixture. I may be finally convinced to relax it!

anniepapa:
@olegpidsadnyi  It will be wonderful! This will not change the fact that reuse non-fixture GIVEN is a bad practice in most cases but still leave flexibility for developers/testers to design their test.

olegpidsadnyi:
@gnikonorov @anniepapa Small update. The given can be called multiple times now. We are heading to the new major release where a lot of limitations would be relaxed. It may introduce some breaking changes as well. Please read the migration section  in the master branch.
The next thing we are researching is the possibility to use gherkin official parser. That would include datatables support perhaps.
Stay tuned

gnikonorov:
Thank you @olegpidsadnyi !

anniepapa:
@olegpidsadnyi : Just tested in my project. Reuse of Given is working like a charm! Look forward to this will be released soon : )

PR #370, opened 2020-06-16
Cleanuptests2

olegpidsadnyi:
Another portion of tests using testdir.

There's one folder left that would require some more creative rewrites using subfolders.


pytest_bdd/cucumber_json.py
```diff
@@ -35,8 +35,8 @@ def add_options(parser):
 
 def configure(config):
     cucumber_json_path = config.option.cucumber_json_path
-    # prevent opening json log on slave nodes (xdist)
-    if cucumber_json_path and not hasattr(config, "slaveinput"):
+    # prevent opening json log on worker nodes (xdist)
+    if cucumber_json_path and not hasattr(config, "workerinput"):
```
youtux, line 39: if this fixes #369, can you mention it in that issue and add a changelog entry?

PR #376, opened 2020-06-25
Given step is no longer a fixture

olegpidsadnyi:
The fixture param is removed from the step definition.
The function of the given step is no longer a fixture. This allows to execute Given multiple times.
Strict gherkin is removed. This would be also up to the official gherkin parser to decide.
Other minor fixes. Moving towards 4.0.0 release


README.rst
```diff
@@ -1222,69 +1159,36 @@ The output will be like:
...
         """I have a custom bar."""
 
 As as side effect, the tool will validate the files for format errors, also some of the logic bugs, for example the
 ordering of the types of the steps.
 
-Migration of your tests from versions 2.x.x
-------------------------------------------------
 
-In version 3.0.0, the fixtures ``pytestbdd_feature_base_dir`` and ``pytestbdd_strict_gherkin`` have been removed.
+Migration of your tests from versions 3.x.x
```
youtux, line 1170: I would leave instructions for migrations for older versions of pytest-bdd.

CHANGES.rst
```diff
@@ -2,7 +2,12 @@ Changelog
 =========
 
 
-- Fixes pytest-xdist 2.0 compatibility (olegpidsadnyi) #369.
+- Strict Gherkin option is removed. (olegpidsadnyi)
+- Given step is no longer a fixture. The scope parameter is also removed. (olegpidsadnyi)
```
youtux, line 6: Let's leave a link to the "how to migrate from v3.0.0" anchor

pytest_bdd/plugin.py
```diff
@@ -104,7 +103,6 @@ def item_key(item):
         else:
             declaration_order = 0
         func, linenum = item.reportinfo()[:2]
-        info = (func, linenum) if linenum is not None else (func, -1)
-        return (info, declaration_order)
+        return (func, linenum if linenum is not None else -1, declaration_order)
```
youtux, line 106: why not "else 0"? "-1" will make items with None suddenly the most important ones.

pytest_bdd/scenario.py
```diff
@@ -111,8 +110,12 @@ def _execute_step_function(request, scenario, step, step_func):
         kw["step_func_args"] = kwargs
 
         request.config.hook.pytest_bdd_before_step_call(**kw)
+        target_fixture = getattr(step_func, "target_fixture", None)
```
youtux, line 113: is this going to allow "when" and "then" steps to have a target_fixture param?

pytest_bdd/steps.py
```diff
@@ -183,7 +134,9 @@ def lazy_step_func():
         if converters:
             step_func.converters = lazy_step_func.converters = converters
 
-        lazy_step_func = pytest.fixture(scope=scope)(lazy_step_func)
+        step_func.target_fixture = lazy_step_func.target_fixture = target_fixture
+
+        lazy_step_func = pytest.fixture()(lazy_step_func)
```
youtux, line 139: Isn't this line still making the step function a fixture? I thought this was one of the main thing to remove.

pytest_bdd/steps.py
```diff
@@ -183,7 +134,9 @@ def lazy_step_func():
...
         setattr(get_caller_module(), get_step_fixture_name(parsed_step_name, step_type), lazy_step_func)
```
youtux, line 140: and IIRC this line is to make pytest discover the fixture

README.rst
olegpidsadnyi: They are quite old by now. Let's not have them in the 4.0.0 branch

olegpidsadnyi: I also changed readthedocs to use stable as default. I think we could stick to the master as a development branch. readthedocs will pick stable tag names.

CHANGES.rst
olegpidsadnyi: done

pytest_bdd/plugin.py
olegpidsadnyi: i think -1 is correct. Lint should be for sure on top of everything, this is the only place where we have no line number.

pytest_bdd/scenario.py
olegpidsadnyi: Only in case if the signature allows it. Currently not the case

pytest_bdd/steps.py
olegpidsadnyi: is it not the step implementation is a fixure and this is how you get them by importing all in the conftest?

pytest_bdd/steps.py
olegpidsadnyi: yes. So you can specify "given foo" in the conftest and override it in your local file

README.rst
youtux: Since readthedocs now shows master, let's indeed merge changes to master

README.rst
```diff
@@ -1192,6 +1194,12 @@ Just normal step declaration with the dependency injection should be used.
         pass
 
 
+Strict gherkin option is removed, so the strict_gherkin parameter can be removed from the scenario decorators
```
youtux, line 1197: We should apply standard formatting to code blocks like strict_gherkin, or pytst_bdd_step_validation_error

pytest_bdd/plugin.py
youtux: pytest-pylint is not the only plugin that defines custom pytest's Items. Anybody can generate arbitrary Items, it's [encouraged](https://docs.pytest.org/en/stable/example/nonpython.html) by pytest itself.
This means that any custom Item that uses None will take the highest priority if pytest-bdd is installed.

I won't comment anymore on this, up to you if you want to fix it or not.

README.rst
olegpidsadnyi: i committed right from the web UI :)

pytest_bdd/scenario.py
```diff
@@ -111,8 +110,12 @@ def _execute_step_function(request, scenario, step, step_func):
...
         # Execute the step.
-        step_func(**kwargs)
+        return_value = step_func(**kwargs)
+        if target_fixture:
+            inject_fixture(request, target_fixture, return_value)
```
jirikuncar, line 117: @olegpidsadnyi the `getfixturevalue` is returning generator instead of a value since the `return_value` is a generator for "yielding" fixture.

```python
@when("a cleanup is necessary", target_fixture="foo")
def foo(request):
    obj = setup()
    yield obj.work()
    obj.clean()
```

PR #386, opened 2020-09-08
Fix performance regression from version 4.0.0

youtux:
https://github.com/pytest-dev/pytest-bdd/pull/384 introduced a performance regression caused by `inspect.stack()`.

To give an idea of the size of the problem, collecting tests from a test suite with ~7000 test is 7x slower than it used to be in pytest 3.2.1.

|              | v3.2.1 | v4.0.0 | v4.0.0 with fix |
|--------------|--------|--------|-----------------|
| collect time | 51s    | 367s   | 43s             |
|              | 1x     | 7.19x  | 0.84x           |

This PR addresses the issue by using lower-level api calls (sys._getframe + inspect.getframeinfo)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Unreleased
+----------
+- Fixed performance regression introduced in 4.0.0 where collection time of tests would take way longer than before. (youtux)
+
+
 4.0.0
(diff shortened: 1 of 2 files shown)


youtux:
Version 4.0.1 is released to PyPI with this fix.

PR #388, opened 2020-09-15
Move Feature, Scenario, Step classes into the parser module

youtux:
I want to have a more clear distinction between the parsing logic (and its classes) and the rest of the plugin, so that I can more efficiently rewrite the parser.


olegpidsadnyi (approved):
looks good

PR #405, opened 2021-02-20
'when' decorator  support target_fixture

bulletRush:
'when' decorator  support target_fixture so that we can check 'when' step result more convenient


1Mark:
I have the same view and agree with this, alternatively, if the repo admins prefer, another approach could be to introduce a separate `when` clause e.g `when_with_target_fixture` if they do not want this to become the norm. But there is definitely a use case here. I raised a similar PR #419 @youtux

PR #406, opened 2021-02-25
Update python and pytest

elchupanebrej:
- Add python 3.9 
- Add pytest 6.1, 6.2
-  Drop support of python 2.7, 3.5

### DIFF
--- a/pytest_bdd/steps.py
+++ b/pytest_bdd/steps.py
@@ -36,7 +36,6 @@ def given_beautiful_article(article):
 """
 
 from __future__ import absolute_import
-import inspect
 
 import pytest
 
@@ -45,24 +44,20 @@ try:
 except ImportError:
     from _pytest import python as pytest_fixtures
 
-from .feature import force_encode
 from .types import GIVEN, WHEN, THEN
 from .parsers import get_parser
 from .utils import get_args, get_caller_module_locals
 
 
-def get_step_fixture_name(name, type_, encoding=None):
+def get_step_fixture_name(name, type_):
     """Get step fixture name.
 
-    :param name: unicode string
+    :param name: string
     :param type: step type
-    :param encoding: encoding
     :return: step fixture name
     :rtype: string
     """
-    return "pytestbdd_{type}_{name}".format(
-        type=type_, name=force_encode(name, **(dict(encoding=encoding) if encoding else {}))
-    )
+    return f"pytestbdd_{type_}_{name}"
 
 
 def given(name, converters=None, target_fixture=None):
@@ -118,7 +113,7 @@ def _step_decorator(step_type, step_name, converters=None, target_fixture=None):
         parser_instance = get_parser(step_name)
         parsed_step_name = parser_instance.name
 
-        step_func.__name__ = force_encode(parsed_step_name)
+        step_func.__name__ = str(parsed_step_name)
 
         def lazy_step_func():
             return step_func
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,6 @@ import re
 
 from setuptools import setup
 
-
 dirname = os.path.dirname(__file__)
 
 long_description = (
@@ -42,8 +41,9 @@ setup(
         "Programming Language :: Python :: 2",
         "Programming Language :: Python :: 3",
     ]
-    + [("Programming Language :: Python :: %s" % x) for x in "2.7 3.5 3.6 3.7 3.8".split()],
-    install_requires=["glob2", "Mako", "parse", "parse_type", "py", "pytest>=4.3", "six>=1.9.0"],
+    + [("Programming Language :: Python :: %s" % x) for x in "3.6 3.7 3.8 3.9".split()],
+    python_requires=">=3.6",
+    install_requires=["glob2", "Mako", "parse", "parse_type", "py", "pytest>=4.3"],
     # the following makes a plugin available to py.test
     entry_points={
         "pytest11": ["pytest-bdd = pytest_bdd.plugin"],
--- a/tests/feature/test_tags.py
+++ b/tests/feature/test_tags.py
@@ -1,6 +1,7 @@
 """Test tags."""
 import textwrap
 
+import pkg_resources
 import pytest
 
 from pytest_bdd.parser import get_tags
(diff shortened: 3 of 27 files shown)


tests/feature/test_tags.py
```diff
@@ -234,7 +237,17 @@ def i_have_baz():
...
+    # Deprecate --strict after pytest 6.1
+    # https://docs.pytest.org/en/stable/deprecations.html#the-strict-command-line-option
+    pytest_version = pkg_resources.get_distribution("pytest").version
+    SemVer = namedtuple("SemVer", ["major", "minor", "patch"])
+    pytest_semver = SemVer(*re.match(r"(\d+)\.(\d+)\.(\d+.*)", pytest_version).groups())
+
+    if int(pytest_semver.major) > 6 or (int(pytest_semver.major) == 6 and int(pytest_semver.minor) >= 2):
+        result = testdir.runpytest_subprocess("--strict-markers")
+    else:
+        result = testdir.runpytest_subprocess("--strict")
```
youtux, line 250: what about
```suggestion
    pytest_version = pkg_resources.get_distribution("pytest").parsed_version
    if pytest_version >= pkg_resources.parse_version("6.2")
        strict_option = "--strict-markers"
    else:
        strict_option = "--strict"
    result = testdir.runpytest_subprocess(strict_option)
```

pytest_bdd/steps.py
```diff
@@ -61,7 +59,7 @@ def get_step_fixture_name(name, type_, encoding=None):
     :rtype: string
     """
     return "pytestbdd_{type}_{name}".format(
-        type=type_, name=force_encode(name, **(dict(encoding=encoding) if encoding else {}))
+        type=type_, name=str(name).encode(**(dict(encoding=encoding) if encoding else {}))
     )
```
youtux, line 63: force_encode(...) was always a noop in python3, but now the code actually does something different, it encodes it to bytes.
I think this should just be:
```suggestion
    return f"pytest_bdd_{type_}_{name}"
```

tox.ini
```diff
@@ -29,6 +31,10 @@ deps =
     -r{toxinidir}/requirements-testing.txt
 commands = {env:_PYTEST_CMD:pytest} {env:_PYTEST_MORE_ARGS:} {posargs:-vvl}
 
+; Black doesn't support >py38 now
 [testenv:py38-pytestlatest-linters]
 deps = black
 commands = black --check --verbose setup.py docs pytest_bdd tests
+
+[testenv:py36-pytestlatest-coverage]
+setenv = PYTHONIOENCODING = utf-8
```
youtux, line 40: what is this for?

setup.py
```diff
@@ -42,8 +42,8 @@
         "Programming Language :: Python :: 2",
         "Programming Language :: Python :: 3",
     ]
-    + [("Programming Language :: Python :: %s" % x) for x in "2.7 3.5 3.6 3.7 3.8".split()],
-    install_requires=["glob2", "Mako", "parse", "parse_type", "py", "pytest>=4.3", "six>=1.9.0"],
+    + [("Programming Language :: Python :: %s" % x) for x in "3.6 3.7 3.8 3.9".split()],
+    install_requires=["glob2", "Mako", "parse", "parse_type", "py", "pytest>=4.3"],
```
youtux, line 46: we should also add a 
```python
python_requires=">=3",
```

Can you also add an entry in the CHANGES.rst file, under a new "Unreleased" release?
Like this:
```
Changelog
=========

Unreleased
-----------
- Drop compatibility for python 2 and officially support only python >= 3.6.

4.0.2
-----
```

youtux (changes requested):

tox.ini
elchupanebrej: There are a few failing tests in the master branch only for py36-pytestlatest-coverage. I found that this environment variable solves the issue. There were a few Unicode fixes in the 3.7 python release, but I didn't investigate deep, because have more prior tasks.

pytest_bdd/steps.py
elchupanebrej: When I took look at this code I've also thought so, but a lot of tests become red. So I've decided to not make refactoring this time but just leave old behavior.

setup.py
elchupanebrej: Thnx

tox.ini
elchupanebrej: Or I could create an issue but seems that priority will be very low and finally it won't be fixed until 3.6 python would be dropped

pytest_bdd/steps.py
youtux: but this is not doing what the old code did. In python 3,`force_encode(obj)` always returned `obj` without changing anything.

Now it is converting the object to bytes, and then we string format that bytes object.

With the proposed change, the result will be like
`"pytest_bdd_given_b'there_is_a_foo'"`
instead of
`"pytest_bdd_given_there_is_a_foo"`

setup.py
```diff
@@ -42,8 +42,9 @@
         "Programming Language :: Python :: 2",
         "Programming Language :: Python :: 3",
     ]
-    + [("Programming Language :: Python :: %s" % x) for x in "2.7 3.5 3.6 3.7 3.8".split()],
-    install_requires=["glob2", "Mako", "parse", "parse_type", "py", "pytest>=4.3", "six>=1.9.0"],
+    + [("Programming Language :: Python :: %s" % x) for x in "3.6 3.7 3.8 3.9".split()],
+    python_requires=">=3.6, <4",
```
youtux, line 46: I wouldn't state that this project is not compatible with python 4. For what we know it is.

tests/feature/test_tags.py
```diff
@@ -1,6 +1,9 @@
 """Test tags."""
+import re
 import textwrap
+from collections import namedtuple
```
youtux, line 4: please remove the unused imports

youtux (changes requested):

tox.ini
elchupanebrej: Removed; Workaroundnadded to failing test

tests/feature/test_tags.py
elchupanebrej: Done

tests/scripts/test_generate.py
```diff
@@ -186,10 +176,12 @@ def test_unicode_characters(testdir):
         ),
     )
 
+    if (3, 6) <= sys.version_info < (3, 7):
+        testdir._pytester._monkeypatch.setenv("PYTHONIOENCODING", "utf-8")
```
youtux, line 180: this is a private variable that is present only in recent versions of pytest. We should avoid using private APIs. Would it work if you would use the `monkeypatch` fixture instead?

```python
def test_unicode_characters(testdir, monkeypatch):
    ...
    if sys.version_info < (3, 7):
        monkeypatch.setenv("PYTHONIOENCODING", "utf-8")
```
I can't reproduce the failure though. (I checked out this PR, removed lines 179-180, run `tox -e py36-pytestlatest-coverage` and it succeeded)

youtux (approved):

youtux:
@elchupanebrej Merged, thank you!

elchupanebrej:
@youtux thanks for great review!

PR #407, opened 2021-03-03
Allow wrap scenarios-generated tests with decorators

elchupanebrej:
- Add scenarios() decorators keyword

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changelog
 Unreleased
 -----------
 - Drop compatibility for python 2 and officially support only python >= 3.6.
+- Allow pass decorators to ``scenarios()`` to wrap every generated test with them.
 
 4.0.2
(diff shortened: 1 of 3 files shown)


youtux:
What's the reason for this feature? If you want to have a fixture automatically used for each scenarion, you can more simply use an autouse fixture:

```python
# test_foo.py

import pytest
from pytest_bdd import scenarios


@pytest.fixture(autouse=True)
def foo_fixture():
    print('Foo was setup')
    yield 'foo'
    print('Foo was teardown')

scenarios("my.feature")
```

elchupanebrej:
My specific case is a complex setup of environment invoking pytest_generate_tests hook and set up some specific dynamic fixtures there. Another case - is specific logging. 

Another thing that thought about was adding just one decorator, but in this case, I have to do the composition of decorators every time

PR #410, opened 2021-03-29
Drop the pytest dot

The-Compiler:
`pytest` has been the recommended spelling (rather than `py.test` ever [since pytest 3.0.0 (2016)](https://docs.pytest.org/en/stable/changelog.html#id1309), and pytest has been dropping the dot in its docs even before that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
-BDD library for the py.test runner
-==================================
+BDD library for the pytest runner
+=================================
 
 .. image:: http://img.shields.io/pypi/v/pytest-bdd.svg
    :target: https://pypi.python.org/pypi/pytest-bdd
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -739,7 +739,7 @@ Note that if you use pytest `--strict` option, all bdd tags mentioned in the fea
 `markers` setting of the `pytest.ini` config. Also for tags please use names which are python-compartible variable
 names, eg starts with a non-number, underscore alphanumberic, etc. That way you can safely use tags for tests filtering.
 
-You can customize how hooks are converted to pytest marks by implementing the
+You can customize how tags are converted to pytest marks by implementing the
```
The-Compiler, line 742: Unrelated but probably not worth a separate PR - converting hooks to marks doesn't make any sense, so this should say "tags" instead.

youtux (approved):
lgtm!

PR #417, opened 2021-04-15
Switch to GitHub actions

youtux:
Disable Travis CI (which is shutting down) and switch over to GitHub actions, as we did for pytest-factoryboy.


youtux:
Github checks is waiting for a notification from Travis CI, but it will never arrive, since this PR disables travis.

olegpidsadnyi (dismissed):

The-Compiler:
FWIW I'm against declaring Python 3.10 support before it's in beta (i.e. no significant changes to Python itself anymore); especially if our testsuite doesn't run with the latest alpha.

youtux:
@The-Compiler 👍

PR #418, opened 2021-04-16
Create and publish PYPI release on tag

youtux:


youtux:
Given the recent security incident with Codecov, I don't feel comfortable with having package publishing done by CI, at least not now.
We don't release pytest-bdd so often that it's worth the risk.
Closing this PR.

PR #419, opened 2021-04-23
Alternative When step decorator.

1Mark:
When asserting the return value of a function it must be passed from
the `when` clause, to the `then` clause.
The cleanest way is to use `target_fixture`.


youtux:
Superseded by #431

PR #425, opened 2021-05-26
step decorators accept re Pattern and parse Parser directly

elchupanebrej:


youtux:
I don't see a reason to have this to be honest.

elchupanebrej:
Some reasons:

- Lowering coupling: give the possibility to user use subclasses of parsers directly without subclassing pytest_bdd parsers
- More clear and narrow API for parsers which pytest_bdd could operate with

Actually, this change is not about "Must" but about "Nice to have" :)

olegpidsadnyi:
When we started with these parsers we made them explicitly passed in order to avoid the ambiguity. It is possible that the patterns in different implementation have similar syntax. It is the way to mark the syntax, not an obstacle really. And the switch case of isinstance is counteracting the OOP we did there.

Issue #409, opened 2021-03-26
Parsers should parse scenario outline like standard scenarios


dcendents:
Hi,

- I use cfparse when I want to parse a list.
- I use scenario outlines when I want to test my scenario with different values

I cannot mix both conditions together.

e.g.:

```
Scenario Outline: mix case
  Given the emails <emails>
  ...

Examples:
  | emails |
  | someone@somewhere.com |
  | someone@somewhere.com, someone.else@somewhere.com |


Scenario: other scenario
  Given the emails email1@server.net, email2@server.net
  ...
```

It would be very useful if I could then have the following step

```python
@given(parsers.cfparse("Given the emails {emails:str+}". extra_types=dict(str=str)), target_fixture="emails")
def emails(emails)
    return emails
```

Ok I could split the list myself, but it's ugly

```python
@given("Given the emails <emails>")
@given(parsers.cfparse("Given the emails {emails:str+}". extra_types=dict(str=str)), target_fixture="emails")
def emails(emails)
    return emails if isinstance(emails, list) else emails.split(",")
```

The same kind of support has been asked a few times in #293 but since it has nothing to do with the original problem I fear it has been lost through all the discussions.

Thanks

TBBle:
We've had to use exactly this workaround of the double-`@given`, once with `<>` examples and once with `parsers.parse`, so implementing this would be great.

I recently worked in [Godog](https://github.com/cucumber/godog) (the official Cucumber implementation for Go) and it works this way; the steps never see the `<example>` placeholder, they parse the resulting _interpolated_ statement like if it was written literally, and for pytest-bdd that would let us apply all the various benefits of `parsers.X` (built-in or custom) as desired.

I haven't checked #433 to see if that's how it worked, but it was reverted in #440 so I guess that's academic.

Issue #404, opened 2021-02-15
Mix scenario outline and parser


haomingyin:
I'm trying to mix scenarios and parser to prevent repeating step definitions.

For example,
```
  Scenario Outline: Create users
    Given I want to create a user
    And my 'given_name' is <given_name>
    And my 'family_name' is <family_name>
    And my 'username' is <username>
    When I create the user
    Then the user will be created

    Examples:
      | given_name | family_name | username |
      | Tom        | Bob         | tom.bob  |
```

and the step definition is 
```
@given("my '{attr_name}' is '<attr_value>'")
def build_request(attr_name, attr_value):
    # I would expect when attr_name is given_name and attr_value will be Tom
    # when attr_name is family_name and attr_value is Bob ...
    pass
```

obviously, it does not work. But I'm wondering if pytest-bdd supports use cases like that. As the user has 10+ attributes, it would be great to not repeat the step definitions.

dcendents:
This already works, I have steps that mix both, but you need to use a parser to parse arguments (one of `parse`, `cfparse` or `re`).

e.g.: 

```python
@given(parsers.parse("my '{attr_name}' is '<attr_value>'"))
def build_request(attr_name, attr_value):
    pass
```

reference: https://github.com/pytest-dev/pytest-bdd#step-arguments

youtux:
This should be fixed by https://github.com/pytest-dev/pytest-bdd/pull/445

Issue #293, opened 2019-03-06
parsers.parse with scenario outline - pytest_bdd.exceptions.StepDefinitionNotFoundError


Vic152:
Hey Guys,

I have a step in scenario outline:

```
Given something
When thing <thing> is moved to position <position>
Then something else

```
```
Examples:
|thing|position|
|1    |1       |
```

In my glue code I write
@when(parsers.parse('thing {thing:d} is moved to position {position:d}'))

I when an error: pytest_bdd.exceptions.StepDefinitionNotFoundError: Step definition is not found: When

I tried When thing {thing} is moved to position {position}

When I use <thing> is passes it as a string. I though if I use parse I could make it an int. Is there some special way feature file has to be written?

sliwinski-milosz:
Hi,

You should define [scenario outlines](https://pytest-bdd.readthedocs.io/en/latest/#scenario-outlines) in following way:
```
Scenario Outline: Some outline
    Given something
    When <thing> is moved to <position>
    Then something else

    Examples:
    | thing | position |
    |   1   |    1     |
```

As you can see `thing` and `position` are in between `<>`

I don't know if you can use scenario outlines together with parsers.
Example in documentation shows another way of specifying converters. In your case:
```
@scenario(
    'yours.feature',
    'Some outline',
    example_converters=dict(thing=int, position=int)
)
def test_outlined():
    pass

...
@when('<thing> is moved to <position>')
def thing_is_moved_to_position(thing, position):
    assert isinstance(thing, int)
    assert isinstance(position, int)
```

Vic152:
Good to know there is a way to do this like below:
>@scenario(
   > 'yours.feature',
   > 'Some outline',
>    example_converters=dict(thing=int, position=int)
>)

I may cast them too... I just thought you could use parser to post-process them 
Thanks a lot!

> As you can see `thing` and `position` are in between `<>`

Sorry formatting in GitHub

Evolter:
Hi, I did get the same error when I was trying to combine "step arguments" with "scenario outlines".
I'm not 100% sure if this is a correct way of doing it, but with `pytest-bdd==3.1.0` I manage to make this working like so:
```feature
  Scenario Outline: Basic DuckDuckGo API Query
    Given the DuckDuckGo API is queried with "<phrase>" using "json" format
    Then the response status code is "200"
    And the response contains results for "<phrase>"

    Examples: Animals
    | phrase   |
    | panda    |
    | python   |
    | platypus |

[...]

PR #433, opened 2021-07-14
Allows use <params> in parsers defined steps

elchupanebrej:
This covers #409 #404 #293

Blocked by #391, it has to be merged first

### DIFF
--- a/pytest_bdd/parser.py
+++ b/pytest_bdd/parser.py
@@ -1,4 +1,3 @@
-import io
 import os.path
 import re
 import textwrap
@@ -457,4 +456,5 @@ def get_tags(line):
     return {tag.lstrip("@") for tag in line.strip().split(" @") if len(tag) > 1}
 
 
-STEP_PARAM_RE = re.compile(r"\<(.+?)\>")
+STEP_PARAM_TEMPLATE = "<{param}>"
+STEP_PARAM_RE = re.compile(STEP_PARAM_TEMPLATE.format(param="((?<=<)[^<>]+(?=>))"))
--- a/pytest_bdd/scenario.py
+++ b/pytest_bdd/scenario.py
@@ -10,15 +10,16 @@ test_publish_article = scenario(
     scenario_name="Publishing the article",
 )
 """
-import collections
 import os
 import re
+from itertools import chain, product
 
 import pytest
 from _pytest.fixtures import FixtureLookupError
 
 from . import exceptions
 from .feature import get_feature, get_features
+from .parser import STEP_PARAM_RE, STEP_PARAM_TEMPLATE
 from .steps import get_step_fixture_name, inject_fixture
 from .utils import CONFIG_STACK, get_args, get_caller_module_locals, get_caller_module_path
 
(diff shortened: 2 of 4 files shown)


elchupanebrej:
@olegpidsadnyi , please review

pytest_bdd/scenario.py
```diff
@@ -10,47 +10,92 @@
...
-                continue
 
-            converters = getattr(fixturedef.func, "converters", {})
-            for arg, value in parser.parse_arguments(name).items():
-                if arg in converters:
-                    value = converters[arg](value)
-                if request:
-                    inject_fixture(request, arg, value)
-            parser_name = get_step_fixture_name(parser.name, type_)
+    parsers_fixturedef_functions = filter(
```
olegpidsadnyi, line 58: Can this use list comprehension syntax? I think it might be a little bit more readable if we don't use map/filter with lambdas.

pytest_bdd/parser.py
```diff
@@ -457,4 +456,5 @@ def get_tags(line):
     return {tag.lstrip("@") for tag in line.strip().split(" @") if len(tag) > 1}
 
 
-STEP_PARAM_RE = re.compile(r"\<(.+?)\>")
+STEP_PARAM_TEMPLATE = "<{param}>"
+STEP_PARAM_RE = re.compile(STEP_PARAM_TEMPLATE.format(param="([^<>]+?)"))
```
olegpidsadnyi, line 460: I wonder what was "?" doing here. Is what between <> even optional?

olegpidsadnyi (changes requested):
Interesting. Can you share the use case where the examples are mixed with the step arguments?

pytest_bdd/parser.py
elchupanebrej: *?, +?, ??
The '*', '+', and '?' qualifiers are all greedy; they match as much text as possible. Sometimes this behaviour isn’t desired; if the RE <.*> is matched against '<a> b <c>', it will match the entire string, and not just '<a>'. Adding ? after the qualifier makes it perform the match in non-greedy or minimal fashion; as few characters as possible will be matched. Using the RE <.*?> will match only '<a>'.

elchupanebrej: For example:
<<nested parameter> partial outer parameter> this regex is matched to "nested parameter". Without ? it would match "nested parameter> partial outer parameter" which is undesired

pytest_bdd/scenario.py
elchupanebrej: Done

olegpidsadnyi (dismissed):
This looks good to me. What do you think, @youtux?

olegpidsadnyi:
@elchupanebrej Does it really need to be dependent on #391? I'm not sure yet about that one. Cucumber uses rule option no-unused-variables for raising errors on that. Is #391 related to your change or you can make it an independent one?

elchupanebrej:
@olegpidsadnyi I'm interested in 391 also. I'll prepare a new patch with adequate API for it and will try to not use it in this patch

elchupanebrej:
@olegpidsadnyi, there is no more dependence on #391

olegpidsadnyi (approved):
This looks okay to me.

olegpidsadnyi:
@elchupanebrej could you make cov happy again?

elchupanebrej:
@olegpidsadnyi done

elchupanebrej:
Just one thing I don't like about this PR: values from Examples section are converted using converters and after that serialized again to str. I would like to have a possibility to substitute Step variables directly, but this would require huge refactoring

olegpidsadnyi:
Sorry, I was having the second thoughts and reverted the merge. I don't really understand why do we need to substitute anything at all in the name of the step. Previously it was just a total match. Otherwise a parser match.
Substitution with the stringified fixture value is a slippery slope. Especially this introduces a lot of additional code that is probably not even needed.
What was the original problem again? Need to escape the <>?

elchupanebrej:
https://github.com/pytest-dev/pytest-bdd/issues/293#issuecomment-494439655
and whole #293

For my project I've solved this by using regular expressions parser and special converter, but I really prefer DRY way.

For this PR I check two approaches - full substitution of variables and approach presented in this PR. The second one is more flexible, but more dangerous

PR #436, opened 2021-07-29
use py3.10 beta 4

graingert:


graingert:
it seems the default branch is failing with https://github.com/pytest-dev/pytest-bdd/pull/426/checks?check_run_id=3194380318#step:4:30

symonk (approved):

PR #439, opened 2021-08-10
Free vars in steps examples

elchupanebrej:
This address #390 (merged and updated #391 )


elchupanebrej:
@olegpidsadnyi @hicksjduk please review the approach to work with free variables in parameters or examples; I've saved behavior to work with parameterized tests from pytest side

pytest_bdd/parser.py
```diff
@@ -281,11 +284,45 @@ def validate(self):
...
+        params = self.params
+        example_params = self.get_example_params()
+        if self.allow_example_free_variables or example_params.issubset(params):
+            return
+        else:
+            raise exceptions.ScenarioExamplesNotValidError(
+                (
+                    """Scenario "{}" in the feature "{}" does not have valid examples. """
+                    """Set of example parameters {} should be a subset of step """
+                    """parameters {} if examples open variables are not allowed"""
```
hicksjduk, line 301: Are we calling them "open" variables, or "free" variables? Does either term have a clear and obvious meaning?

hicksjduk (commented):
Looks good, just one comment about terminology.

CHANGES.rst
```diff
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 -----------
+- Add options to control free variables in Examples and step definitions
```
olegpidsadnyi, line 6: yeah.. what's free? is it all about checking the unused variables or not?

olegpidsadnyi:
As for the options it would be annoying to specify this on each scenario. It is easier to turn this option off in the .ini file once and for all. See https://docs.pytest.org/en/6.2.x/writing_plugins.html#using-hooks-in-pytest-addoption - there should be something about pytest plugin options.

auvipy (dismissed):

CHANGES.rst
elchupanebrej: It's just about terminology) Free == unused
https://en.wikipedia.org/wiki/Free_variables_and_bound_variables

elchupanebrej:
> As for the options it would be annoying to specify this on each scenario. It is easier to turn this option off in the .ini file once and for all. See https://docs.pytest.org/en/6.2.x/writing_plugins.html#using-hooks-in-pytest-addoption - there should be something about pytest plugin options.

Thanks! Hadn't take this into view

hicksjduk:
Thanks guys, excellent ideas.

elchupanebrej:
Added options to control such behavior from ini files

olegpidsadnyi:
@elchupanebrej  The paradigm has changed based on our latest discussion. The unification of the <> and {} as you suggested lead to the more cucumber way of looking at it. So there are no <params> like this anymore. Just like in cucumber it is just a step template that substitutes the <params> with example values before attempting to match the step and it may match with the {pattern} of the step.
These params are no longer fixtures either. This is quite a change and this validation of the unused param doesn't make much sense anymore.
```
params = frozenset(sum((list(step.params) for step in self.steps), []))
```
I'm not sure if Step.params is any useful anymore.
I think we should just remove this particular validation and related tests like `test_wrongly_outlined` without introducing any option. Let's make it less restrictive. I don't see how validation helps in practice.

elchupanebrej:
During the implementation of #478 I've reviewed this approach and found that in case if we have outlined scenarios we could use extra columns for joining example tables as part of a primary key. To make this feature unambiguous I propose that both joinable tables have to include such key. In another case just skip parameters set as ambiguous

Also, I found that sometimes impossible to combine Example tables if they are unbalanced by parameter count so some parameters could be unfilled. In such case, I propose to not render parameters. This case could be intercepted by a specific step definition

For tag columns I propose to use "@" sign to mark the column as a tag column, so:

```gherkin
Feature: Outline
    Examples:
    | fruits    | @     | @      | key | # | status |  
    | apples    | red   | fresh  |  1  | # |   ill  |
    | oranges   | green | rotten |  2  | # | hungry |

    Examples:
    | fruits  | # | status |
    | peaches | # |  full  |

    Scenario Outline: Outlined given, when, thens
        Given there are <start> <fruits>
        When I eat <eat> <fruits>
        Then I should have <left> <fruits>
        # And I am <status>

        Examples:
        | start | eat | left | key |
        |   12  |  5  |  7   |  2  |
        |    5  |  4  |  1   |  1  |

        Examples:
        | start | eat | left |
        |   16  |  9  |  7   |
        |    8  |  3  |  5   |
        |    9  |  7  |  2   |
```
So in this example will be executed:
2 scenarios: 1 for apples and 1 for oranges, because they are combined by "key", and they have additional tags
3 scenarios for peaches

"status" column is just commented but still there

olegpidsadnyi:
If we want to remove the restriction that the variables in the example table should match the steps we should focus only on that.

olegpidsadnyi:
I can't find any recommendations on the official gherkin reference how to treat the unused table variables.
I can imagine you want to comment out one step for debugging reasons that would make a table variable unused. I don't want to make it an option anymore. Let's just remove this validation logic and leave the validation to gherkin linter in your editor or command line tools.

PR #441, opened 2021-08-22
Update readme accordingly

olegpidsadnyi:
Updated readme the usage of the example tables and the migration section.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,6 @@ Scenario decorator
 The scenario decorator can accept the following optional keyword arguments:
 
 * ``encoding`` - decode content of feature file in specific encoding. UTF-8 is default.
-* ``example_converters`` - mapping to pass functions to convert example values provided in feature files.
 
 Functions decorated with the `scenario` decorator behave like a normal test function,
 and they will be executed after all scenario steps.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -112,7 +112,6 @@ Scenario decorator
 The scenario decorator can accept the following optional keyword arguments:
 
 * ``encoding`` - decode content of feature file in specific encoding. UTF-8 is default.
```
youtux, line 114: let's remove this too? I don't think anybody uses any other encoding

README.rst
```diff
@@ -1208,37 +1206,27 @@ As as side effect, the tool will validate the files for format errors, also some
 ordering of the types of the steps.
 
 
-.. _Migration from 3.x.x:
+.. _Migration from 4.x.x:
```
youtux, line 1209: I would leave the old migration section, not everybody updates to each version of the library at every major release

README.rst
```diff
@@ -1208,37 +1206,27 @@ As as side effect, the tool will validate the files for format errors, also some
...
 
-Migration of your tests from versions 3.x.x
+Migration of your tests from versions 4.x.x
 -------------------------------------------
 
-
-Given steps are no longer fixtures. In case it is needed to make given step setup a fixture
-the target_fixture parameter should be used.
-
+The templated steps should use step argument parsers in order to match the scenario outlines
```
youtux, line 1214: we should also mention that the values of the examples will not be injected as fixtures anymore

README.rst
olegpidsadnyi: Okay.. Leaving the code part intact if you want to change it in your branch.

README.rst
youtux: @olegpidsadnyi ⬆️

PR #445, opened 2021-09-18
Rewrite scenario/feature examples logic

youtux:
Rewrite the logic to parse Examples for Scenario Outlines. Now the substitution of the examples is done during the parsing of Gherkin feature files.

We won't need to define the steps twice like ``@given("there are <start> cucumbers")`` and ``@given(parsers.parse("there are {start} cucumbers"))``. The latter will be enough.

Supersedes https://github.com/pytest-dev/pytest-bdd/pull/433.
Fixes https://github.com/pytest-dev/pytest-bdd/issues/409, https://github.com/pytest-dev/pytest-bdd/issues/293


olegpidsadnyi (approved):
Nice, a lot more consistent this way

Issue #448, opened 2021-09-28
Test steps fixture parametrization is broken by #445


elchupanebrej:
After #445 document paragraph "Combine scenario outline and pytest parametrization" became obsolete

tests/feature/test_parametrized.py introduces a new approach (parts of template language are put into steps):
```
Feature: Parametrized scenario
        Scenario: Parametrized given, when, thens
             Given there are {start} cucumbers
             When I eat {eat} cucumbers
             Then I should have {left} cucumbers
```

elchupanebrej:
After #439 there could be a possibility to get such behavior by implementing non-rendered steps and using usual pytest fixtures

PR #455, opened 2021-10-24
Fix combining of scenario outline and pytest parametrization after #445

elchupanebrej:
This address to #448

### DIFF
--- a/pytest_bdd/parser.py
+++ b/pytest_bdd/parser.py
@@ -4,6 +4,8 @@ import textwrap
 import typing
 from collections import OrderedDict
 
+from _pytest.fixtures import FixtureRequest
+
 from . import exceptions, types
 
 SPLIT_LINE_RE = re.compile(r"(?<!\\)\|")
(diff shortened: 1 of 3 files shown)


bbatliner:
This is exactly the patch I was expecting to see after encountering `KeyErrors` related to the parameterization example in the README with the `<start>` syntax.

However, Oleg's [comment](https://github.com/pytest-dev/pytest-bdd/issues/448#issuecomment-951179675) on the related issue suggests that this feature will be removed in the future?

Am I understanding this correctly? Why would this be removed?

elchupanebrej:
Yes, this feature will be removed. An approach where a test is parametrized inside python coda and these parameters are used in Gherkin scenario definition is broken "by design" because hides some information that has to be present in Examples, so you couldn't share your scenario with stakeholders without losing some information needed for scenario execution.
I found that there is some backward incompatibility and created this patch. But with approval from core maintainers, I'll close this patch after removing rudiments of old approach

elchupanebrej:
Superseded by #469

PR #457, opened 2021-10-30
Monospace font for file names in READNE

adamatan:


youtux:
It doesn't look like you achieved the result you mention though:

Screenshot 2021-11-14 at 18 19 50

Maybe we should instead include the file name in a comment at the beginning of the snippet instead.

Issue #466, opened 2021-12-17
New test failures with pytest 7.0.0rc1


musicinmybrain:
Fedora Linux is testing pytest 7.0.0rc1 with existing Python packages, and this revealed some new test failures in `pytest-bdd` with pytest 7. These were bisected to https://github.com/pytest-dev/pytest/commit/6d6cfd839aa850061ddef65922eddb7567280e5f by @hroncok.

```
=================================== FAILURES ===================================
________________________ test_feature_path_not_found[.] ________________________

testdir = <Testdir local('/tmp/pytest-of-mockbuild/pytest-0/test_feature_path_not_found0')>
base_dir = '.'

    @pytest.mark.parametrize("base_dir", NOT_EXISTING_FEATURE_PATHS)
    def test_feature_path_not_found(testdir, base_dir):
        """Test feature base dir."""
        prepare_testdir(testdir, base_dir)

        result = testdir.runpytest("-k", "test_not_found_by_ini")
>       result.assert_outcomes(passed=2)
E       AssertionError: assert {'deselected': 8,\n 'errors': 0,\n 'failed': 0,\n 'passed': 2,\n 'skipped': 0,\n 'warnings': 0,\n 'xfailed': 0,\n 'xpassed': 0} == {'deselected': 0,\n 'errors': 0,\n 'failed': 0,\n 'passed': 2,\n 'skipped': 0,\n 'warnings': 0,\n 'xfailed': 0,\n 'xpassed': 0}
E         Common items:
E         {'errors': 0,
E          'failed': 0,
E          'passed': 2,
E          'skipped': 0,
E          'warnings': 0,
E          'xfailed': 0,
E          'xpassed': 0}
E         Differing items:
E         {'deselected': 8} != {'deselected': 0}
E         Full diff:
E           {
E         -  'deselected': 0,
E         ?                ^
E         +  'deselected': 8,
E         ?                ^
E            'errors': 0,
E            'failed': 0,
E            'passed': 2,
E            'skipped': 0,
E            'warnings': 0,
E            'xfailed': 0,
E            'xpassed': 0,
E           }

base_dir   = '.'
result     = <RunResult ret=ExitCode.OK len(stdout.lines)=9 len(stderr.lines)=0 duration=0.05s>
testdir    = <Testdir local('/tmp/pytest-of-mockbuild/pytest-0/test_feature_path_not_found0')>

[...]

PR #470, opened 2022-01-02
Fix "deselect" keyword usage by pytester.assert_outcomes for pytest>6

elchupanebrej:
This address #466

### DIFF
--- a/tests/feature/test_feature_base_dir.py
+++ b/tests/feature/test_feature_base_dir.py
@@ -1,7 +1,10 @@
 """Test feature base dir."""
+import pkg_resources
 import pytest
 
 NOT_EXISTING_FEATURE_PATHS = [".", "/does/not/exist/"]
+
+assert_outcomes_extra = pkg_resources.get_distribution("pytest").parsed_version >= pkg_resources.parse_version("7.0.0")
 
 
 @pytest.mark.parametrize("base_dir", NOT_EXISTING_FEATURE_PATHS)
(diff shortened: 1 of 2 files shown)


elchupanebrej:
Let wait until pytest 7.0 release

tests/feature/test_feature_base_dir.py
```diff
@@ -1,24 +1,33 @@
 """Test feature base dir."""
+import pkg_resources
 import pytest
 
 NOT_EXISTING_FEATURE_PATHS = [".", "/does/not/exist/"]
 
+assert_outcomes_extra = pkg_resources.get_distribution("pytest").parsed_version >= pkg_resources.parse_version("7.0.0")
```
The-Compiler, line 7: Note that `pkg_resources` is a third-party dependency (part of `setuptools`, but not the stdlib), and is [officially discouraged](https://setuptools.pypa.io/en/latest/pkg_resources.html):

> Use of pkg_resources is discouraged in favor of importlib.resources, importlib.metadata, and their backports (resources, metadata). Please consider using those libraries instead of pkg_resources.

Note that `importlib.metadata` also only was added in Python 3.8. However, pytest 7 added [`pytest.version_tuple`](https://docs.pytest.org/en/7.0.x/reference/reference.html#version-tuple), so you could get the version from there (and assume `False` when it does not exist).

The-Compiler (changes requested):
FWIW I opened a discussion about this on the pytest side too: https://github.com/pytest-dev/pytest/issues/9471

If we decide to stick with the change (likely, I think), having a single `assert_outcomes` helper function might lead to cleaner code than having if's scattered everywhere.

Also note that `pkg_resources` is discouraged, I added a comment about it.

elchupanebrej:
Closed because of #466

PR #469, opened 2022-01-01
Refuse combining of scenario outline and pytest parametrization

elchupanebrej:
This address #448
Supersedes #455

PR #472, opened 2022-01-03
Default parser switched to cfparse

elchupanebrej:
In most cases, cfparse parser is more useful than a string parser. It could be used almost everywhere, just one edge-case is when the step definition contains "{....}" pair. In this case user still could select string parser


elchupanebrej:
Have to merged after #469 because it brokes obsolete test case

PR #473, opened 2022-01-03
Refactoring to have possibility fix few defects

elchupanebrej:


elchupanebrej:
Have to be merged after #469 because it brokes obsolete test case

Issue #476, opened 2022-01-04
print tags in gherkin-terminal-reporter


guillego:
Hello, I am using the Gherkin tags in the feature specification to refer scenarios to different software requirements and it would be really convenient to be able to get those tags printed when I run pytest with the `--gherkin-terminal-reporter` option. I know they are generated into the cucumber JSON but that's less human readable as of right now.

The feature looks like this:
` @R1.1 @R1.2
  Scenario: Send packages to socket
    Given a service listening on a random socket
    When tool sends "2" packages to socket
    Then "2" packages are received in the service`

So it would be great to get the tags formatted in some way in the output of the test:

$ pytest --gherkin-terminal-reporter -vv
===================================================================================== test session starts =====================================================================================                                                                          

tests/pacsio_test.py::test_send_packages_to_socket <- utilities/pacs_io/.venv/lib/python3.8/site-packages/pytest_bdd/scenario.py 
Feature: Tool for network testing
    Scenario: Send packages to socket
    Given a service listening on a random socket
    When tool sends "2" packages to socket
    Then "2" packages are received in the service
    PASSED
    (tags: R1.1, R1.2)

guillego:
I'm trying to implement it as a feature (first-time contributor to OSS), but I'm having trouble with the test, if I define a test-case for it like this in `feature/test_gherkin_terminal_reporter.py`:

```
FEATURE_WITH_TAGS = """\
@feature_tag0 @feature_tag1
Feature: Gherkin terminal output feature
    @scenario_tag0 @scenario_tag1
    Scenario: Scenario example 1
        Given there is a bar
        When the bar is accessed
        Then world explodes
"""
```

Then pytest throws this:

`E   pytest.PytestUnknownMarkWarning: Unknown pytest.mark.scenario_tag1 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html`

guillego:
I can't find a contributing guideline but I have something that works in my forked branch (minus the testing, which I can't get to properly work). How does the PR process work?

ryux1:
I have an implementation ready that prints scenario tags in both `-v` and `-vv` Gherkin terminal output, with coverage for both modes. I’ll open the PR now.

ryux1:
Correction: I found the existing #807 and #477 after checking the complete open-PR list, so I closed my duplicate #830. Sorry for overlooking those PRs in the initial issue-reference search.

PR #477, opened 2022-01-04
Modify gherkin terminal reporter to print tags

guillego:
As mentioned in #476, this PR implements a simple parser for printing the tags in the verbose gherkin terminal reporter.

### DIFF
--- a/pytest_bdd/gherkin_terminal_reporter.py
+++ b/pytest_bdd/gherkin_terminal_reporter.py
@@ -36,6 +36,10 @@ class GherkinTerminalReporter(TerminalReporter):
     def __init__(self, config):
         super().__init__(config)
 
+    def _write_tags(self, tags: list, prefix: str = ''):
+        tags_string = ', '.join(tags)
+        return prefix + '(' + tags_string + ')'
+
     def pytest_runtest_logreport(self, report):
         rep = report
         res = self.config.hook.pytest_report_teststatus(report=rep, config=self.config)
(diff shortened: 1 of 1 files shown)


pytest_bdd/gherkin_terminal_reporter.py
```diff
@@ -75,9 +80,13 @@ def pytest_runtest_logreport(self, report):
         elif self.verbosity > 1:
             if hasattr(report, "scenario"):
                 self.ensure_newline()
+                self._tw.write(self._write_tags(report.scenario["feature"]["tags"]), **tag_markup)
+                self._tw.write("\n")
                 self._tw.write("Feature: ", **feature_markup)
                 self._tw.write(report.scenario["feature"]["name"], **feature_markup)
                 self._tw.write("\n")
+                self._tw.write(self._write_tags(prefix="    ", tags=report.scenario["tags"]), **tag_markup)
```
arthuRHD, line 88: It might be better to use `\t`instead of the space bar for a tabulation.

pytest_bdd/gherkin_terminal_reporter.py
```diff
@@ -75,9 +80,13 @@ def pytest_runtest_logreport(self, report):
...
+                self._tw.write("\n")
                 self._tw.write("    Scenario: ", **scenario_markup)
```
arthuRHD, line 90: Same here.

arthuRHD (changes requested):

youtux:
Hi @guillego, this looks like something we would like to merge, but it would needs some tests to be added to the PR.

Issue #378, opened 2020-07-08
Use official parser


jirikuncar:
Is there an interest in switching to the official Gherkin parser: https://github.com/cucumber/cucumber/tree/master/gherkin/python? 

I would be willing to prepare a PR.

Cheers!

The-Compiler:
As far as I know, @olegpidsadnyi started some work on this (or at least planned to) - see e.g. https://github.com/pytest-dev/pytest-bdd/pull/366#issuecomment-648014646

anniepapa:
@The-Compiler @olegpidsadnyi  Do you have any idea when will be the next release of Pytest-bdd? I was asked twice already from company guild meeting about this. Very much looking forward to it.  :smile:

youtux:
I actually abandoned the idea (for the moment) of using the parser from cucumber. I am trying instead to rewrite the parser.
You can see the progress in the [rewrite-parser branch](https://github.com/pytest-dev/pytest-bdd/tree/rewrite-parser).

elchupanebrej:
@youtux could you please provide some insights into why rewriting was selected over using the official parser?

Maybe it's possible to create some intermediate adapter layer(or model) to make possible support a few different parsers?
Seems that pytest-bdd Gherkin dialect and official Gherkin have some independent features. pytest-bdd has much less community than Gherkin so it is vitally necessary to support official features and just after that support extensions. Seems that it's needed to give some integration space inside. So next evolution steps could be:

- Separate intermediate adapter model (and prove that legacy parser could work with it)
- Develop adapter model to connect model provided by the official parser and intermediate model 
- Extend official parser to support features provided by pytest-bdd
- Extend pytest-bdd run model to support features from the official Gherkin dialect through the adapter model

TBBle:
It looks like this has become the blocker for Datatables, i.e. https://github.com/pytest-dev/pytest-bdd/pull/366#issuecomment-648014646, but the [rewrite-parser branch](https://github.com/pytest-dev/pytest-bdd/tree/rewrite-parser) also stalled in September 2020, over twelve months ago.

Is there still seen to be a way forward in this direction (replacing the parser, either a rewrite or integrating the Gherkin python parser), or do we need to resuscitate #366 against current pytest-bdd master in order to advance that support?

(Or perhaps point people at external addons like [sttable](https://github.com/pytest-dev/pytest-bdd/pull/180#issuecomment-557793431) although I don't know how well-maintained that is, as I haven't actually tried it.)

[...]

PR #489, opened 2022-01-12
Preparing the scenario for multiple example tables

olegpidsadnyi:
Cucumber has added multiple example tables to the scenarios. These examples can also get tags.
We need to support pytest marks for those examples.
The parser is going to add multiple example tables to the TemplatedScenario. The tags can become part of the Examples object.
Since each test item is produced by the parametrization represented by a single table row we have to pass the marks parameter to each of them.

### DIFF
--- a/pytest_bdd/parser.py
+++ b/pytest_bdd/parser.py
@@ -228,9 +228,14 @@ class ScenarioTemplate:
         self.feature = feature
         self.name = name
         self._steps: typing.List[Step] = []
-        self.examples = Examples()
+        self.example_tables = [Examples()]
         self.line_number = line_number
         self.tags = tags or set()
+
+    @property
+    def examples(self):
+        """Current examples."""
+        return self.example_tables[-1]
 
     def add_step(self, step):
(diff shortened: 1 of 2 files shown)


olegpidsadnyi:
@elchupanebrej have a look. I started sketching out some preparations for the multiple scenario examples

blaisep:
@olegpidsadnyi , 
https://github.com/pytest-dev/pytest-bdd/issues/378#issuecomment-2040791260

ChernyavskiyAlex:
Hi @youtux, is there any chance that something similar will be delivered within 8.0.0 release?

youtux:
@jsa34 do we support this out of the box now that we are using the official parser?

jsa34:
Ooh good question... I'll take a look

ChernyavskiyAlex:
@jsa34 according to my observations gherkin parser provides tags but they are not taking into consideration on pytest-bdd side when execution "from_dict" methods

jsa34:
Superseded by #738

PR #498, opened 2022-02-01
Update black dependency

elchupanebrej:
Black was released a stable version.


youtux:
Already in master now, but thanks!

PR #503, opened 2022-02-22
Allow "yield" in step definitions

### LINKED ISSUE #392: @given with target_fixture does not behave like pytest-fixture on yield anymore
In v4.0.1 "given"-steps with "target_fixture" being set do no longer behave the same way pytest-fixtures do when a value is yielded. This is either a not explicitly documented change in the behaviour since 3.4.0 or a bug.

When using such a "given"-step, I do no longer get the yielded value (as I would expect from a pytest-fixture), but a generator. This is a hindrance for e.g. using context managers within a given-step that ensure that something will be cleaned up at the end of a test case.

Let me give you one example:

```py
@given("a running server", target_fixture="running_server")
def run_server():
    with subprocess.Popen("./start_my_server",  stdout=subprocess.PIPE) as proc:
        yield proc

@when("I do something")
def do_something():
   # e.g. send some request to the server

@then("the server does something")
def the_server_does_something(running_server):
    assert running_server.stdout.read() == "some kind of information"
    # running_server now is a generator in v4.0.1 but has been a subprocess.Popen in 3.4.0
```

youtux:
This way step definitions can define code to be executed during teardown of the test item.
Apparently this behaviour was broken by the 4.0 release.

Fixes #392.

### DIFF
--- a/tests/feature/test_steps.py
+++ b/tests/feature/test_steps.py
@@ -484,3 +484,51 @@ def test_step_trace(testdir):
     result.assert_outcomes(failed=1)
     result.stdout.fnmatch_lines(["*test_when_step_validation_error*FAILED"])
     assert "INTERNALERROR" not in result.stdout.str()
+
+
+def test_steps_with_yield(testdir):
+    """Test that steps definition containing a yield statement work the same way as
+    pytest fixture do, that is the code after the yield is executed during teardown."""
+
+    testdir.makefile(
+        ".feature",
+        a="""\
+Feature: A feature
+
+    Scenario: A scenario
+        When I setup stuff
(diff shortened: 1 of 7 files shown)


tests/feature/test_steps.py
```diff
@@ -484,3 +484,51 @@ def test_when_step_validation_error():
...
+            """\
+        import pytest
+        from pytest_bdd import given, when, then, scenarios
+
+        scenarios("a.feature")
+
+        @when("I setup stuff", target_fixture="stuff")
+        def stuff():
+            print("Setting up...")
+            yield 42
```
bharath-basepair, line 514: yield is not supporting in given

PR #509, opened 2022-03-14
Don't make steps fixtures

therve:
This is an attempt at removing how steps are registered via fixtures,
and instead use a custom registry. It seems that using fixtures for that
is a bit awkward, as we can't differentiate steps from regular fixtures
apart from the parser attribute, and we need to access private pytest
attributes to do so. Given that pytest_bdd both produces and consumes
the steps, it would seem that we can remove it. As an added bonus it
should have a minor performance increase.

This is mostly a proof of concept, as I wanted to start the discussion.
Let me know if it doesn't make sense, thanks.

### DIFF
--- a/pytest_bdd/generation.py
+++ b/pytest_bdd/generation.py
@@ -9,7 +9,7 @@ import py
 from mako.lookup import TemplateLookup
 
 from .feature import get_features
-from .scenario import find_argumented_step_fixture_name, make_python_docstring, make_python_name, make_string_literal
+from .scenario import make_python_docstring, make_python_name, make_string_literal
 from .steps import get_step_fixture_name
 from .types import STEP_TYPES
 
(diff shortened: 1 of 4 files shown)


youtux:
I like the idea, but I wonder if that works with how fixtures are expected to be overridden by each module.
I wonder if we even have a test for that, but in any case we should support that.

In the doc we only mention that steps defined by the parent packages can be used by the children packages (https://pytest-bdd.readthedocs.io/en/latest/#reusing-steps).
But if a step is defined by a child package, it should stay within the scope of that package, and not be available to parent packages.

Does it make sense to you what I mean?

therve:
Ah that makes sense. I never used that but it would seem that this change breaks this capability, and customizing the registry would make things too complicated and duplicate fixture purposes. I'll think of other smaller changes, thanks!

PR #521, opened 2022-05-27
README : start_cucumbers -> cucumbers

NickVeld:
`start_cucumbers` is not defined anywhere but there is the fixture named `cucumbers`


youtux:
Thanks for reporting this, fixed in https://github.com/pytest-dev/pytest-bdd/pull/526

PR #539, opened 2022-07-12
Fix `parsers.re` doing only partial string match

youtux:
This PR make sure that `parsers.re` does a fullmatch of the re, rather than ignoring non-matching characters at the end of the string.
All the other parsers already work this way, there is no reason for `parsers.re` not to.

Checklist

- [x] Merge https://github.com/pytest-dev/pytest-bdd/pull/534 first


The-Compiler:
FWIW this broke a couple of tests for me where I used trailing periods: https://github.com/qutebrowser/qutebrowser/commit/418f91dfcd2955e32721eb718c278b38696fd1d5

Is that intended? I'm guessing yes, but I have no idea how other Gherkin parsers handle this.

youtux:
It is expected to break a few steps indeed, I will release a new major version for this.

maybe I’ll mention in the release notes how to migrate to this, although I assume that if you use regex, you know how to consume all characters at the end of the string.

PR #541, opened 2022-07-16
Rewrite docs to not use meaningful function names for decorated step functions

youtux:
Continuation of the work started in https://github.com/pytest-dev/pytest-bdd/pull/536.

Having the documentation using only `_` for step function names will encourage users to not give meaningful names either.


youtux:
Oleg prefers having meaningful names, so that it's not "surprising" to the user the fact that we use underscore as function name. Closing this.

PR #544, opened 2022-07-23
Fix step function resolution logic

### LINKED ISSUE #542: Reusable step functions regression: Specific step doesn't override generic one anymore
After #534, with this `conftest.py`:

```python
import pytest_bdd as bdd
import pytest

@pytest.fixture
def value():
    return []

@bdd.when(bdd.parsers.parse("I have a {thing} thing"))
def generic(thing, value):
    value.append(thing)
```

this `test_specific.py`:

```python
import pytest_bdd as bdd

@bdd.when("I have a specific thing")
def specific(value):
    value.append("42")

@bdd.then(bdd.parsers.parse("The value should be {thing}"))
def check(thing, value):
    assert value == [thing]


bdd.scenarios("specific.feature")
```

and this `specific.feature`:

```feature
Scenario: Overlapping steps 1
        When I have a specific thing
        Then the value should be 42

Scenario: Overlapping steps 2
        When I have a generic thing
        Then the value should be generic
```

I would expect that `specific` takes precedence over `generic`, i.e., the value for the first test is `"42"`, not `"generic"`. This used to be the case, but isn't anymore after that commit:

```
    @bdd.then(bdd.parsers.parse("The value should be {thing}"))
    def check(thing, value):
>       assert value == [thing]
E       AssertionError: assert ['specific'] == ['42']
E         At index 0 diff: 'specific' != '42'
E         Use -v to get more diff
```

Note, however, it *does* work fine when `generic` is moved from the `conftest.py` file into `test_specific.py`:

youtux:
The step function resolution logic has long been broken. The refactor in #534 made this bug more visible, since it treats all step definitions with and without parsers as the same (previously, definitions without parsers had higher priority).

Also, we now require pytest>=6.2 (I needed to use some newer pytest APIs).

Fixes #542.


youtux:
@The-Compiler would you feel like reviewing the [part that uses pytest's internals](https://github.com/pytest-dev/pytest-bdd/pull/544/files#diff-bedfbbde3664dab01f0c672fd9a33236698fce6eb32a3d59276de855d20d3d5b)?

PR #570, opened 2022-11-03
Drop dependency on glob2

drothlis:
The standard library's `glob` supports "**" since Python 3.5. pytest-bdd requires Python >= 3.7.

TODO:

- [x] Merge #571 and rebase.


drothlis:
CI failure against `pytest-latest`:

    INTERNALERROR>     tw = py.io.TerminalWriter()
    INTERNALERROR> AttributeError: module 'py' has no attribute 'io'

It's unrelated to this pull request. pytest 7.2.0 (2022-10-23) [release notes] say:

> [#10396](https://github.com/pytest-dev/pytest/issues/10396): pytest no longer depends on the `py` library. pytest provides a vendored copy of `py.error` and `py.path` modules but will use the `py` library if it is installed. If you need other `py.*` modules, continue to install the deprecated `py` library separately, otherwise it can usually be removed as a dependency.

[release notes]: https://docs.pytest.org/en/stable/changelog.html#pytest-7-2-0-2022-10-23

youtux (approved):

PR #571, opened 2022-11-03
Drop dependency on py.io

drothlis:
Since pytest 7.2.0 was released 2 weeks ago (2022-10-23), CI fails against `pytest-latest`:

    INTERNALERROR>     tw = py.io.TerminalWriter()
    INTERNALERROR> AttributeError: module 'py' has no attribute 'io'

Pytest 7.2.0 dropped its dependency on the `py` package. We don't specify `py` explicitly as a dependency, so tests starting failing since the release of pytest 7.2.0.

To drop this dependency, pytest vendored `py.io.TerminalWriter` into `_pytest._io` in pytest commit [276405a03] (6.0.0rc1). We require pytest>=6.2, so we can import it from there.

[276405a03]: https://github.com/pytest-dev/pytest/commit/276405a03

### DIFF
--- a/src/pytest_bdd/generation.py
+++ b/src/pytest_bdd/generation.py
@@ -5,8 +5,12 @@ import itertools
 import os.path
 from typing import TYPE_CHECKING, cast
 
-import py
 from mako.lookup import TemplateLookup
+
+try:
+    from _pytest._io import TerminalWriter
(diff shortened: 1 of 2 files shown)


youtux (approved):
Thank you!

PR #572, opened 2022-11-04
tests: Use `pytester` fixture instead of `testdir`

drothlis:
`testdir` is deprecated. It's identical to `pytester`, except that `testdir` returns `py.path` objects whereas `pytester` returns Python stdlib `Path` objects. So I had to update usages of those paths.

`pytester` was added in pytest 6.2, which is our minimum supported version.


tests/feature/test_feature_base_dir.py
```diff
@@ -5,44 +5,44 @@
...
+    prepare_pytester(pytester, base_dir)
 
-    result = testdir.runpytest("-k", "test_ok_by_param")
+    result = pytester.runpytest("-k", "test_ok_by_param")
     result.assert_outcomes(passed=2)
 
 
-def prepare_testdir(testdir, ini_base_dir):
-    testdir.makeini(
+def prepare_pytester(pytester, ini_base_dir):
```
youtux, line 44: This should still be called prepare_testdir

tests/library/test_parent.py
```diff
@@ -87,7 +87,7 @@ def _():
         )
     )
 
-    testdir.mkpydir("subdir").join("test_global_when.py").write(
+    pytester.mkpydir("subdir").joinpath("test_global_when.py").write_text(
```
youtux, line 90: I think we can use the `/` operator if now it's an instance of pathlib.Path
(Not important though)

youtux (changes requested):
Thank you for this! Just a minor function rename, the rest is good for me!

tests/feature/test_feature_base_dir.py
drothlis: Done.

drothlis:
> This should still be called prepare_testdir

@youtux I put it back to `prepare_testdir`. I'm not sure why the CI checks aren't running, maybe because I squashed & force-pushed?

youtux:
Oh please don't force push, it will make it a pain for me to review changes since my last review

drothlis:
@youtux sorry about that! If you click on this link it'll show you the changes:

image

https://github.com/pytest-dev/pytest-bdd/compare/0d73b7cabcdfc84fc517f247e99489936eecd397..c8677a7783c4373a0ff21be8aea5eb899d16e1a5

drothlis:
Summary of API changes (in case anyone needs to fix conflicts with older branches):

old | new
---|---
`testdir.tmpdir` | `pytester.path`
py.path `join()` | pathlib.Path `joinpath()`
py.path `basename` | pathlib.Path `name`
py.path `dirname` | pathlib.Path `parent`
py.path `relto()` | pathlib.Path `relative_to()`
py.path `.strpath` | `str(<pathlib.Path>)`
py.path `read()` | pathlib.Path `read_text()`
py.path `write()` | pathlib.Path `write_text()`
py.path.write's `ensure=True` | pathlib.Path: Call `mkdir()` before calling `write_text()`

youtux (approved):

Issue #607, opened 2023-04-04
Bug: Parser Arguments being overwritten


samuelspagl:
Hey there, 

first of all thanks a lot for this amazing project, I really love working with it. :) 

### Describe the situation

I'm trying to reuse my fixtures / step definitions as often as possible to make my test code concise and clean. That's why I rely a lot on the feature of using multiple BDD decorators for one function.
In one situation, which is for my case not really uncommon, I want to use a function like this:

```python
@given("Do some stuff", target_fixture="stuff")
@when(parsers.parse("you did a \"{action_type}\"", target_fixture="stuff")
def do_some_stuff(action_type: str = None):
   ...
```

The argument (action_type) for the second decorator (when) will never be correctly filled, but always uses the fallback value `None`

### Expected Behaviour

The BDD argument should be correctly inserted as an argument for the function, and the "fallback" value of the function should only be used if there's no input for this argument. 

### Test system: 
💻: *MacOS Ventura*
🐍: *3.10*
Pytest Version: *7.2.2*
Pytest_BDD Version: *6.1.1*

I also tested it with a fresh install of pytest and pytest_bdd

Thanks a lot in advance

jsa34:
Fixed in #737

PR #610, opened 2023-04-18
fix step arguments with default value

samuelspagl:
The bug is mentioned in the issue #607.
This is a sample implementation trying minimal invasive by implementing a new function that retrieves all arguments with default values.

Another solution would be to let `get_args` retrieve a `dict` with the arguments as keys and its values as values. But this would add two or three more changes I think, therefore I didn't implement this solution. 

I'm really looking forward hearing from you. 

Best
Samuel


yunxuo:
why not fix now?

yunxuo (approved):
test pass

PR #617, opened 2023-04-29
Support asterisks as continuation steps on par with And and But

Artalus:
[Gherkin Documentation mentions](https://cucumber.io/docs/gherkin/reference/#Asterisk) you can uses `*` (asterisk) to describe lists of things, same way as with `And` steps:
```gherkin
Scenario: All done
  Given I am out shopping
  * I have eggs
  * I have milk
  * I have butter
  When I check my list
  Then I don't need anything
```
Dunno if anybody actually needs such feature - since search for `asterisk` did not yield anything neither in PRs nor Issues. But since the change is literally 1 line of code, I thought I might propose it just as well and see what happens :smile: 

Not sure if the place for testing it is right (`tests/feature/test_steps.py`) or if it should be done someplace else tho.

### DIFF
--- a/src/pytest_bdd/parser.py
+++ b/src/pytest_bdd/parser.py
@@ -26,6 +26,7 @@ STEP_PREFIXES = [
     # Continuation of the previously mentioned step type
     ("And ", None),
     ("But ", None),
+    ("* ", None),
 ]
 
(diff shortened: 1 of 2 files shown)


jsa34:
@youtux can we close this as it is now done and covered by a test in the test suite?

jsa34:
@Artalus that's fab, thank you!

Yes, indeed - I also added a separate PR to add a test to cover this to ensure it is meeting the same outcome as this PR

PR #623, opened 2023-07-06
improve parser performance

dcendents:
Hi,

Using py-spy to check where to improve the performance of my tests, I noticed a lot of time was spent in the Step name property.
Just by caching the value instead of recalculating it every time I gain ~15%.
Same gain when I run the tests in parallel

Sequence:
* Before:
    * 1280 passed, 49 deselected in 125.52s (0:02:05)
* After:
    * 1280 passed, 49 deselected in 107.67s (0:01:47)

Parallel:
* Before:
    * 1280 passed in 60.61s (0:01:00)
* After:
    * 1280 passed in 50.24s

It is a very minor change, hope it can be merged.

Thanks


youtux (changes requested):
hi, interesting find. I think it may have been already a cached property in the past, but the for some reason I must have removed it.

Could you change the implementation by using `functools.cached_property` instead? It should result in a way neater implementation.

Also, make sure to invalidate the cached attribute when `self.lines` or `self._name` are modified.

src/pytest_bdd/parser.py
```diff
@@ -334,9 +337,15 @@ def name(self) -> str:
...
 
+    @property
+    def name(self) -> str:
+        return self.full_name
+
     @name.setter
     def name(self, value: str) -> None:
         self._name = value
+        if "full_name" in self.__dict__:
+            del self.__dict__["full_name"]
```
youtux, line 348:
```suggestion
        if "full_name" in self.__dict__:
            del self.full_name
```

src/pytest_bdd/parser.py
```diff
@@ -318,9 +319,11 @@ def add_line(self, line: str) -> None:
         :param str line: Line of text - the continuation of the step name.
         """
         self.lines.append(line)
+        if "full_name" in self.__dict__:
+            del self.__dict__["full_name"]
```
youtux, line 323:
```suggestion
            del self.full_name
```

youtux (changes requested):

src/pytest_bdd/parser.py
dcendents: I did try that first but it broke all the tests.

From https://stackoverflow.com/questions/59899732/python-cached-property-how-to-delete

"Your code will try to delete the method itself, and not the cached value. The correct way to do it is to delete the key from self.__dict__ (which is where the data is actually stored)."

src/pytest_bdd/parser.py
dcendents: I did try that first but it broke all the tests.

From https://stackoverflow.com/questions/59899732/python-cached-property-how-to-delete

"Your code will try to delete the method itself, and not the cached value. The correct way to do it is to delete the key from self.__dict__ (which is where the data is actually stored)."

dcendents:
Hi @youtux ,

Actually you were right we can `del self.full_name` but we do need to validate the key is in `__dict__` first, which I was not doing the first time around and caused the error I mentionned.

Anyway that showed my modifications were lacking proper testing so I added a unit test as well to validate the cache behavior.
I really tried to use a mock to wrap the step object and spy on the `full_name` method (property) but I could not make it work, so I hope you are ok with the way I tested it.

Changing values directly on the step object to show the cache value is still returned (which should never happen in production code).
Changing values using properties show the cache is invalidated.

Cheers

src/pytest_bdd/parser.py
dcendents: fixed it

src/pytest_bdd/parser.py
dcendents: fixed it

youtux (approved):

Issue #362, opened 2020-05-21
error when using with --testmon


jnoortheen:
I have bdd tests and trying to use `pytest --testmon` from https://pypi.org/project/pytest-testmon/

But it fails with 
```
 ❯ pytest --testmon                                                                                                                                  
Exit: The stored data file /home/noor/src/src/py/divisible/yolo/.testmondata version (4) is not compatible with current version (6). You must delete the stored data to continue.
Traceback (most recent call last):
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/bin/pytest", line 10, in <module>
    sys.exit(main())
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/_pytest/config/__init__.py", line 124, in main
    ret = config.hook.pytest_cmdline_main(
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/noor/.virtualenvs/yolo-fc-KERI9mRf-py3.8/lib/python3.8/site-packages/_pytest/main.py", line 240, in pytest_cmdline_main
    return wrap_session(config, _main)
[...]

PR #641, opened 2023-10-23
Issue #362: Only unconfigure if configured

n3world:
`pytest_unconfigure` can be called even if `pytest_configure` was not. In order to avoid errors during unconfigure only perform the unconfigure if the configure was performed. Add a counter which is incremented before configure to track the number of times pytest started initialization so that the config is not popped early from the stack.

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -135,3 +135,21 @@ def test_pytest_bdd_after_scenario_called_after_scenario(pytester):
 
     assert scenario_1.name == "Scenario 1"
     assert scenario_2.name == "Scenario 2"
+
+
+def test_pytest_unconfigure_without_configure(pytester):
+    """
+    Simulate a plugin forcing an exit during configuration before bdd is configured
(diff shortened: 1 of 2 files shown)


tests/test_hooks.py
```diff
@@ -135,3 +135,17 @@ def pytest_bdd_after_scenario(request, feature, scenario):
 
     assert scenario_1.name == "Scenario 1"
     assert scenario_2.name == "Scenario 2"
+
+
+def test_pytest_unconfigure_without_configure(pytester):
+    pytester.makeconftest(
```
youtux, line 141: please add a docstring explaining that this can happen in case plugins cause pytest to exit early

youtux (changes requested):

tests/test_hooks.py
n3world: Added docstring

n3world:
I also resolved the unit test failures from previous run

youtux (approved):

PR #646, opened 2023-11-01
fix(typing): improve decorator type hinting

JP-Ellis:
The type hinting for the most commonly used decorators were incomplete, resulting in decorated functions being obscured.

This makes use of a `TypeVar` which allows the return type of the decorator to be the same as its input.

I have also taken the opportunity to fix other instances of `Callable` type hints missing their arguments.

---

PS:

The benefit of moving `import` statement behind the `if TYPE_CHECKING` block is to avoid unnecessary parsing of libraries at runtime (when no type checking takes place). There were a few cases of:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Callable
```

which provides no benefit as the `typing` library has to be parsed in the first place anyway.

I hope it's alright that I fixed a couple of these :)

### DIFF
--- a/src/pytest_bdd/plugin.py
+++ b/src/pytest_bdd/plugin.py
@@ -1,16 +1,15 @@
 """Pytest plugin entry point. Used for any fixtures needed."""
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Callable, cast
+from typing import TYPE_CHECKING, Any, Callable, Generator, TypeVar, cast
 
 import pytest
+from typing_extensions import ParamSpec
 
 from . import cucumber_json, generation, gherkin_terminal_reporter, given, reporting, then, when
 from .utils import CONFIG_STACK
 
 if TYPE_CHECKING:
-    from typing import Any, Generator
-
     from _pytest.config import Config, PytestPluginManager
     from _pytest.config.argparsing import Parser
     from _pytest.fixtures import FixtureRequest
@@ -19,6 +18,10 @@ if TYPE_CHECKING:
     from pluggy._result import _Result
 
     from .parser import Feature, Scenario, Step
+
+
+P = ParamSpec("P")
+T = TypeVar("T")
 
 
 def pytest_addhooks(pluginmanager: PytestPluginManager) -> None:
@@ -93,7 +96,7 @@ def pytest_bdd_step_error(
     feature: Feature,
     scenario: Scenario,
     step: Step,
-    step_func: Callable,
+    step_func: Callable[..., Any],
     step_func_args: dict,
     exception: Exception,
 ) -> None:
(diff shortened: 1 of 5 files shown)


src/pytest_bdd/plugin.py
```diff
@@ -123,7 +127,7 @@ def pytest_cmdline_main(config: Config) -> int | None:
     return generation.cmdline_main(config)
 
 
-def pytest_bdd_apply_tag(tag: str, function: Callable) -> Callable:
+def pytest_bdd_apply_tag(tag: str, function: TCallable) -> TCallable:
```
youtux, line 130: not sure about this, shouldn't we instead use ParamSpec?

Like
```python
P = ParamSpec("P")
T = TypeVar("T")

def pytest_bdd_apply_tag(tag: str, function: Callable[P, T]) -> Callable[P, T]:
```

src/pytest_bdd/scenario.py
```diff
@@ -197,14 +196,14 @@ def _execute_scenario(feature: Feature, scenario: Scenario, request: FixtureRequ
 
 def _get_scenario_decorator(
     feature: Feature, feature_name: str, templated_scenario: ScenarioTemplate, scenario_name: str
-) -> Callable[[Callable], Callable]:
+) -> Callable[[TCallable], TCallable]:
```
youtux, line 199: ditto here, I think ParamSpec is really what we should use: https://docs.python.org/3.11/library/typing.html#typing.ParamSpec

src/pytest_bdd/steps.py
```diff
@@ -74,10 +74,10 @@ def get_step_fixture_name(step: Step) -> str:
 
 def given(
     name: str | StepParser,
-    converters: dict[str, Callable] | None = None,
+    converters: dict[str, Callable[[Any], Any]] | None = None,
```
youtux, line 77: ditto

youtux (changes requested):

JP-Ellis:
While I agree completely that `ParamSpec` is the preferred option as it gives further introspection to the parameter types, I opted not to use that as `ParamSpec` was introduced in Python 3.10, and therefore won't work for Python 3.8 and 3.9 which `pytest-bdd` supports.

Given you are using `typing_extensions` already, I'll modify the PR to import `ParamSpec` from `typing_extensions` instead of the standard library.

src/pytest_bdd/steps.py
JP-Ellis: I'm not sure whether `Callable[P, T]` is what is needed in this case. What would that imply for:

```python
@given(
    "Some step",
    converts={
        "id": int,
        "valid": bool,
        "list": lambda s: s.split(", "),
    },
)
```

Having a single `Callable[P, T]` would imply that they all have the same return type which is clearly incorrect.

Having said that, it might be correct to say that this should be refined to `Callable[[str], Any]`, unless you are aware of cases where something else than a string is passed to the convert?

youtux: ah right, indeed it should be `Callable[[str], Any]`

youtux (approved):

PR #658, opened 2023-12-02
Fix typing

youtux:
Fix many `mypy` issues.
Notably, I replace private attributes like `__scenario__`, `__scenario_report__`, etc. with registries (WeakKeyDictionary). E.g.
```python
# Old way

# setting value
scenario_wrapper.__scenario__ = scenario

# getting value
scenario = getattr(obj, "__scenario__", None)


# New way
# setting value
scenario_wrapper_template_registry[scenario_wrapper] = scenario

# getting value
scenario = scenario_wrapper_template_registry.get(scenario_wrapper)
```

Other changes:
* Replaced usages of `Any` with a more specific type if possible, otherwise `object`. `Any` causes the type checker to completely forego the type checking. Using `object` is a better way.


jsa34:
@youtux I think mypy is happy now since we merged #698 ?  If it is, perhaps we can close this?

youtux:
Well, mypy is happy only because we put a lot of `#type: ignore`, ideally we wouldn't need those

jsa34:
Good point - so the focus of this would be to eliminate those where possible :)

jsa34 (approved):

PR #660, opened 2023-12-03
Reduce codebase complexity

youtux:


jsa34:
This looks mainly parser-focussed, which is no longer relevant.  Can we just make it about the cast and then we can open further PRs for any codebase improvements we find? :)

youtux:
👍 closing this and making the fix for `setdefault` in https://github.com/pytest-dev/pytest-bdd/pull/730

PR #669, opened 2024-01-21
Fix pytest version check

youtux:


The-Compiler (approved):
Sorry for the radio silence - this seems to fix things from what I can see!

If you'd be willing to lift the minimum pytest version to 7 ([released](https://docs.pytest.org/en/latest/changelog.html#pytest-7-0-0-2022-02-03) almost two years ago), you could also use [`pytest.version_tuple`](https://docs.pytest.org/en/latest/reference/reference.html#pytest-version-tuple) instead. That'd mean not introducing the new dependency on `packaging`.

Or given that you're only interested in major/minor version, you'd probably get away with `tuple(int(c) for c in pytest.__version__.split(".")[:2])` on pytest 6.x (oh, the irony of having a version-dependent version check...).

youtux:
yes I could update to pytest 7, but I don't have much time so I'll just merge this

PR #698, opened 2024-09-05
Implement the official gherkin parser

jsa34:
Implementation of the official Gherkin parser.
Trying to maintain the public API and current codebase as much as possible.  I was thinking little steps, and introduce new features/refactoring after we know it works like-for-like with current feature set. 

Credit to @elchupanebrej for his implementation I referenced in pytest-bdd-ng

**Breaking changes**
Some of the previous features that are not Gherkin-compliant I had to remove. I have tried to find workarounds where possible.

Changes include:
- multiline steps not using triple quotes no longer work
- tags cannot have spaces in them

NB: This simply replaces the parsing logic with the Gherkin official parser, but the current parser does some additional logic and makes custom entities that are needed for test execution.  I have coerced the Gherkin Document data to the existing model (minimising change), but I have raised a new MR (https://github.com/pytest-dev/pytest-bdd/pull/699) that completely removes the existing parser to replace with Gherkin Document data generated.  Not sure which is best route to take.


jsa34:
Finished tinkering and fixing things for the CI jobs :)

jsa34:
@sourcery-ai review

pyproject.toml
```diff
@@ -42,6 +42,8 @@ parse-type = "*"
 pytest = ">=6.2.0"
 typing-extensions = "*"
 packaging = "*"
+gherkin-official = "^29.0.0"
+pydantic = "^2.9.0"
```
youtux, line 46: Id really prefer if we can avoid using pydantic. Its design mixes up model with represantion layer, which is not something I like.

Plus it uses the word `validate` where it should actually say `deserialize`/`load`/`unmarshal`/… — I find it a poorly design API

youtux: It’d be great if we can just use data classes, or if that’s not enough, `attrs`.

src/pytest_bdd/gherkin_parser.py
```diff
@@ -0,0 +1,170 @@
...
+    text: str
+    dataTable: Optional[DataTable] = None
+    docString: Optional[DocString] = None
+
+    @field_validator("keyword", mode="before")
+    def normalize_keyword(cls, value: str) -> str:
+        return value.lower().strip()
+
+    @property
+    def given_when_then(self) -> str:
```
youtux, line 67: What’s this about?

src/pytest_bdd/gherkin_parser.py
```diff
@@ -0,0 +1,170 @@
...
+def _compute_given_when_then(steps: List[Step]) -> List[Step]:
+    last_gwt = None
+    for step in steps:
+        if step.keyword in STEP_TYPES:
+            last_gwt = step.keyword
+        step.given_when_then = last_gwt
+    return steps
+
+
+class GherkinParser:
```
youtux, line 152: Not sure why this is a class, since it has no state. Can’t it just be a function that returns the GherkinDocument?

tests/feature/test_outline.py
```diff
@@ -217,6 +217,6 @@ def _(string):
         r"bork |",
         r"bork||bork",
         r"|",
-        r"bork      \\",
-        r"bork    \\|",
+        "bork      \\",
```
youtux, line 220: This shouldn’t have needed to change

tests/feature/test_steps.py
```diff
@@ -489,19 +495,27 @@ def test_step_trace(pytester):
         from pytest_bdd import given, when, scenario
 
         @given('I have a bar')
-        def _():
+        def i_have_bar():
```
youtux, line 498: Why? I want to encourage users to define steps functions using underscore, since they are not supposed to be invoked directly, and since the name is actually given by the `name` Parma in the decorator

tests/feature/test_steps.py
```diff
@@ -489,19 +495,27 @@ def test_step_trace(pytester):
...
+        def when_it_fails():
             raise Exception('when fails')
 
-        @scenario('test.feature', 'When step has failure')
-        def test_when_fails_inline():
+        @pytest.fixture
+        def dependency():
+            raise Exception('dependency fails')
+
+        @when("it's dependency fails")
```
youtux, line 509: Typo

youtux (changes requested):
Good stuff! A few comments, but we are in the right track! Thank you for your effort

jsa34:
I'll take a look into your comments.

Most I can already see I agree with - I was in 2 minds about Pydantic - it's made model validation much easier when ingesting the gherkin document dict, but I can absolutely see your point.  I will have a look :)

src/pytest_bdd/gherkin_parser.py
jsa34: This is the "Given/When/Then" value of the step, calculated, as the step may be "And", for example.

I think it's referred to in the parser as "type", which I found confusing, especially as type is used to reference other things other than "given/when/then" type of the step.

Happy to use the same name as the parser currently does of "type"

src/pytest_bdd/gherkin_parser.py
jsa34: Sorry - some habits die hard from Java background!  You are absolutely right :)

pyproject.toml
jsa34: Re-implemented with attrs and factory methods for initialise the dataclasses from dicts

tests/feature/test_outline.py
jsa34: > I had to add a converter to the parser

jsa34:
@youtux resolved most comments - just a couple of things:
* given_when_then comments' response -> just need a steer 
* the escaping I have fixed the way I could best see, but would appreciate a second pair of eyes.

tests/feature/test_steps.py
jsa34: This grammatical error exists in the existing test - I have fixed it and the other references, as well as a few other grammatical things I found along the way.

pyproject.toml
jsa34: Apologies - I misread your comment and thought you said "but that's not enough" and so used attrs.  I think dataclasses are fine.  I will commit with removing attr and just using pure dataclasses

src/pytest_bdd/gherkin_parser.py
youtux: yeah I'd find it clear if it was named `type`.
Maybe we should finally make an Enum for the possible values of step types (given, when, then) and return that instead of a mere string

src/pytest_bdd/gherkin_parser.py
```diff
@@ -1,56 +1,94 @@
...
 
 
-class Location(BaseModel):
+@dataclass
+class Location:
     column: int
     line: int
 
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "Location":
```
youtux, line 20: let's use `from __future__ import annotations`, so that here we can do:

```suggestion
    def from_dict(cls, data: Dict[str, Any]) -> Self:
```

src/pytest_bdd/gherkin_parser.py
youtux: let's try to build the Step object only when we know its concrete type, so that we don't need to set it after its instantiation

youtux: actually, do we have already this information in the `keyword` and `keywordType`? I guess we will have an `"And"` possible value, but that's fine for this class, since this class is just a wrapper around the untyped dicts returned by the gherkin-parser lib.

So I'd say to remove this property, and instead determine the correct step type (give/when/then) only when we create an instance of the Step (the business model one, `pytest_bdd.parser.Step`)

youtux (commented):
can you also add a changelog entry in `CHANGES.rst` with all the backwards incompatible changes?
So far I could see that we don't accept spaces in tags and that now `Feature:` is required in the feature file definitions.

tests/feature/test_wrong.py
```diff
@@ -50,4 +50,4 @@ def test_wrong():
     )
     result = pytester.runpytest()
     result.assert_outcomes(errors=1)
-    result.stdout.fnmatch_lines("*FeatureError: Multiple features are not allowed in a single feature file.*")
+    result.stdout.fnmatch_lines("*FeatureError: *")
```
youtux, line 53: what kind of errors would we show to the user in this case? I think it's important to make sure we return a decent error message

jsa34: Ah, the error messages are not nice (from a human readability perspective).

I forgot to go back and now just throw the raw error message.  I am just adding an error message parser to map the parser errors to throw the same errors are before, and a few more that weren't previously covered.

tests/feature/test_outline.py
```diff
@@ -171,7 +171,7 @@ def test_outline_with_escaped_pipes(pytester):
     pytester.makefile(
         ".feature",
         outline=textwrap.dedent(
-            r"""\
+            r"""
```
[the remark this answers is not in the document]
jsa34: My mistake!

src/pytest_bdd/gherkin_parser.py
jsa34: For the moment, I've put in the business logic.  Let me know if you have other thoughts (I just want to focus on the parser here, and avoiding making as many changes outside the parser as possible, but I had already thought along the lines of replacing with an enum - great mind think alike!)

src/pytest_bdd/gherkin_parser.py
```diff
@@ -276,11 +316,30 @@ def get_gherkin_document(abs_filename: str = None, encoding: str = "utf-8") -> G
...
+        gherkin_error_handler = GherkinParserErrorHandler()
+        gherkin_error_handler(message, line, line_content, filename)
+        # If no patterns matched, raise a generic GherkinParserError
+        raise exceptions.GherkinParseError(f"Unknown parsing error: {message}", line, line_content, filename)
+
+    # At this point, the `gherkin_data` should be valid if no exception was raised
     return GherkinDocument.from_dict(gherkin_data)
+
+
+class GherkinParserErrorHandler:
```
youtux, line 332: this can just be a function right?

tests/feature/test_background.py
```diff
@@ -2,16 +2,16 @@
 
 import textwrap
 
-FEATURE = '''\
+FEATURE = """\
```
youtux, line 5: why this change? I prefer the previous version where we didn't need to `\` escape chars inside this heredoc

src/pytest_bdd/gherkin_parser.py
```diff
@@ -276,11 +316,30 @@ def get_gherkin_document(abs_filename: str = None, encoding: str = "utf-8") -> G
...
+        """Map the error message to a specific exception type and raise it."""
+        # Split the raw_error into individual lines
+        error_lines = raw_error.splitlines()
+
+        # Check each line against all error patterns
+        for error_line in error_lines:
+            for pattern, exception_class, message in ERROR_PATTERNS:
+                if pattern.search(error_line):
+                    # If a match is found, raise the corresponding exception with the formatted message
+                    raise exception_class(message, line, line_content, filename)
```
youtux, line 345: we should `raise ... from e` (the original exception), so that the stacktrace is represented correctly, while now it is represented as
```
...
E   (5:5): expected: #EOF, #TableRow, #DocStringSeparator, #StepLine, #TagLine, #ScenarioLine, #RuleLine, #Comment, #Empty, got 'Background:'

During handling of the above exception, another exception occurred:
...
E   pytest_bdd.exceptions.BackgroundError: BackgroundError: Multiple 'Background' sections detected. Only one 'Background' is allowed per feature.
E   Line number: 5
E   Line:     Background:
E   File: /private/var/folders/dy/_0849zr93fj1td2_07t4ybpr0000gn/T/pytest-of-youtux/pytest-4/test_background_basic0/background.feature
```

src/pytest_bdd/gherkin_parser.py
```diff
@@ -8,7 +12,68 @@
 from gherkin.token_scanner import TokenScanner
 
 from . import exceptions
-from .types import STEP_TYPES
+
+if typing.TYPE_CHECKING:
+    from typing import Self
+
+
+ERROR_PATTERNS = [
```
youtux, line 20: absolutely appreciate the effort you put to represent the original exception! I see though that some of them are not tested, would you mind add simple tests for them?
Of course, tests didn't exist before either, so if you don't feel like it it's perfectly fine

tests/feature/test_outline.py
```diff
@@ -171,8 +171,7 @@ def test_outline_with_escaped_pipes(pytester):
     pytester.makefile(
         ".feature",
         outline=textwrap.dedent(
-            r"""
-            Feature: Outline With Special characters
+            r"""Feature: Outline With Special characters
```
youtux, line 174: was this change necessary? it should pass with the original version too

tests/parser/test_parser.py
```diff
@@ -0,0 +1,11 @@
...
+
+from src.pytest_bdd.gherkin_parser import get_gherkin_document
+
+
+def test_parser():
+    test_dir = Path(__file__).parent
+    feature_file = test_dir / "test.feature"
+    feature_file_path = str(feature_file.resolve())
+
+    get_gherkin_document(feature_file_path)
```
youtux, line 11: it'd be great if we can assert the whole structure of this gherkin document, to make sure it is parsed correctly

youtux:
I just tried parsing this feature:
```
Feature: Background support

    Scenario: Basic usage
        And foo

```
 and it fails with an unhandled exception:
```
============================= test session starts ==============================
platform darwin -- Python 3.12.2, pytest-8.3.2, pluggy-1.5.0
rootdir: /private/var/folders/dy/_0849zr93fj1td2_07t4ybpr0000gn/T/pytest-of-youtux/pytest-5/test_background_basic0
plugins: bdd-7.2.0, xdist-3.6.1
collected 1 item

test_background_basic.py F                                               [100%]

=================================== FAILURES ===================================
_______________________________ test_background ________________________________

self = <[AttributeError("'Step' object has no attribute 'lines'") raised in repr()] Step object at 0x107bd4e30>

    def __str__(self) -> str:
        """Return a string representation of the step.
    
        Returns:
            str: A string representation of the step.
        """
>       return f'{self.type.capitalize()} "{self.name}"'
E       AttributeError: 'NoneType' object has no attribute 'capitalize'

/Users/youtux/Developer/pytest-bdd/src/pytest_bdd/parser.py:255: AttributeError
=========================== short test summary info ============================
FAILED test_background_basic.py::test_background - AttributeError: 'NoneType'...
============================== 1 failed in 0.07s ===============================
```

can you fix this and maybe add a test for it?

youtux:
by the way, that was caught running `mypy`, you might want to run that too to find possible other errors

src/pytest_bdd/parser.py
```diff
@@ -301,119 +231,225 @@ class Step:
...
+        )
+        for step in self.parse_steps(scenario_data.steps):
+            scenario.add_step(step)
 
-    :return: List of tags.
-    """
-    if not line or not line.strip().startswith("@"):
-        return set()
-    return {tag.lstrip("@") for tag in line.strip().split(" @") if len(tag) > 1}
+        for example_data in scenario_data.examples:
```
youtux, line 402: scenario_data.examples may be None

src/pytest_bdd/gherkin_parser.py
```diff
@@ -0,0 +1,345 @@
...
+def _to_raw_string(normal_string: str) -> str:
+    return normal_string.replace("\\", "\\\\")
+
+
+def get_gherkin_document(abs_filename: str, encoding: str = "utf-8") -> GherkinDocument:
+    with open(abs_filename, encoding=encoding) as f:
+        feature_file_text = f.read()
+
+    try:
+        gherkin_data = Parser().parse(TokenScanner(feature_file_text))
```
youtux, line 317: do we need to use TokenScanner? from the gherkin README.md, it seems we can just do 
```
parser = Parser()
gherkin_document = parser.parse("Feature: ...")
```

tests/feature/test_outline.py
youtux: aaaah I see the issue now, it's a raw string, so ok nevermind we can keep it as it is

src/pytest_bdd/gherkin_parser.py
jsa34: Good point.  Updated :)

src/pytest_bdd/parser.py
jsa34: The new parser always has an empty list and never None.  Should I coerce to None in the case on an empty list, or keep the parser output as is?

src/pytest_bdd/gherkin_parser.py
jsa34: Fixed and added test

tests/parser/test_parser.py
jsa34: Added to test assertion

src/pytest_bdd/gherkin_parser.py
jsa34: Added tests.  Some I realise don't happen as I expected in the parser, so updated exceptions list and ensured at least one test for each.

src/pytest_bdd/parser.py
youtux: ah then let's remove `| None` from the annotation

src/pytest_bdd/gherkin_parser.py
jsa34: Yep - I started making it more complicated than need be, but the current logic makes sense to be a function.  Updated.

tests/feature/test_background.py
jsa34: Sorry, it was my overzealous replacing from fixing another test.  This was unintentional.  Teaches me to be careful with find and replace!

src/pytest_bdd/parser.py
jsa34: Done in the Gherkin parser :)

youtux (approved):

youtux:
Congrats! We are now using the official parser!

jsa34:
Thank you, @youtux 🙌

PR #699, opened 2024-09-08
Only use the official Gherkin parser objects and remove custom parser logic

jsa34:
Going a step further than my previous PR that simply parsed the feature files using the official parser, but then copied this data into the existing parser objects, this PR removes the old parser completely and either uses the data directly that is parsed into the models from the Gherkin parser or adds needed functionality to the Gherkin parser models to enhance just being DAOs for pytest-bdd to use the data as it needs to.


jsa34:
I'm thinking a different approach might be better to keep the DAO/mapping and logic encapsulated

youtux (commented):
let's first get https://github.com/pytest-dev/pytest-bdd/pull/698 merged, then let's focus on this.

Actually, could you rebase this on top of https://github.com/pytest-dev/pytest-bdd/pull/698, so that in GitHub I can check the diff from that only?

jsa34:
Closed to focus on parser, and ended up just dumping things I was trying on this branch - apologies!

Issue #703, opened 2024-09-21
Tag is missing for the 7.3.0 release


ionenwks:
Seen 7.3.0 was released on pypi but tag seems to be missing here (would use the pypi tarball, but it lacks tests). Thanks.

youtux:
yep done, sorry. One day I will implement in CI this release process

youtux:
why do you need tests in the distribution though?

webknjaz:
@youtux the tests should be included in sdists but with the wheels, it depends on the repository structure. They are necessary for the downstreams to be able to rely on sdists when repackaging.

webknjaz:
You should probably fill out the GitHub Releases entries for the tags, additionally to pushing to Git: https://github.com/pytest-dev/pytest-bdd/releases

webknjaz:
> One day I will implement in CI this release process

Here's my recent state of the art release automation example: https://github.com/ansible/awx-plugins/blob/e22c150/.github/workflows/ci-cd.yml#L748-L1150. And my PyPUG guide with a simpler example: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/. Hope this helps!

youtux:
> @youtux the tests should be included in sdists but with the wheels

is that documented/recommended anywhere by the PyPA or PyPI?

webknjaz:
> the tests should be included in sdists 

To facilitate this, I even made an action that allows testing from sdist in upstream CIs: https://github.com/marketplace/actions/checkout-python-sdist

webknjaz:
> > @youtux the tests should be included in sdists but with the wheels
> 
> is that documented/recommended anywhere by the PyPA or PyPI?

It's a recurring debate which boils down to whether individual upstream maintainers are willing to make downstream-friendly sdists. There's polarized opinions based on this distinction. The recent discussion I can recall was on dpo... I don't remember if we have that over on PyPUG, there were talks about documenting it on the issue tracker, though. Also, modern setuptools auto-include a number of discovered directories into sdists, especially when you start using PEP 621 metadata definition method. If they aren't, add them to `MANIFEST.in`. I usually rely on the `setuptools-scm` plugin which auto-includes everything tracked by Git...

The only thing to be careful about is that the top-level tests dirs should not end up in wheels, because they would end up in the root of site-packages. This is not such a problem for projects that have tests nested into the main installed package dir as they'll be scoped properly.

[...]

PR #704, opened 2024-09-21
Include tests in source distribution

youtux:
As requested in https://github.com/pytest-dev/pytest-bdd/issues/703#issuecomment-2365320563

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,30 @@ on:
   workflow_dispatch:
 
 jobs:
+  build:
+    name: Build distribution 📦
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+      - name: Install pypa/build
+        run: >-
+          python3 -m
+          pip install
+          build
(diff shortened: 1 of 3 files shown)


webknjaz:
The acceptance criteria is that if you unpack the sdist, you should be able to run the tests, just like from a Git checkout.

This essentially means including all the related/mandatory automation bits, like the tox config.

Here's an example of a package that's good:
https://inspector.pypi.io/project/awx-plugins-core/0.0.1a4/packages/76/ea/3232f77020de845b36845ef5a350d8b45def28419314e7661709d9968993/awx_plugins_core-0.0.1a4.tar.gz/

webknjaz:
A practical way to ensuring it's usable, would be integrating my action: https://github.com/marketplace/actions/checkout-python-sdist. With that, you'd build the dists in CI before running the tests. And then retrieve those dists and execute the tests from them, using the action instead of `actions/checkout`. An additional benefit could be installing the wheel to run the tests against it. This also facilitates publishing what's tested if you go for adding release automation later on.

With this, you'll see immediately if what's in an sdist is enough to run the tests/build the docs downstream without forcing them to resort to using Git.

youtux:
https://github.com/pytest-dev/pytest-bdd/issues/703#issuecomment-2377891636

PR #707, opened 2024-09-26
Add CI action for publishing artifacts to PyPI

youtux:


webknjaz:
@youtux it is highly discussed to build the dists in the same job as publishing when you are using the tokenless method. It should happen in a separate job, otherwise you're opening an attack surface with privilege escalation in non-PyPI systems through dependency poisoning. I'm pretty sure I posted this warning in all the guides — did I miss some places?

webknjaz:
Besides, it's good to smoke-test that the dists build well in PRs and run a `twine check --strict` on them continuously..

youtux:
> I'm pretty sure I posted this warning in all the guides — did I miss some places?

Can you point me to this docs?

webknjaz:
@youtux its at the bottom of
https://docs.pypi.org/trusted-publishers/security-model/. And the action readme has it in one of the notes here: https://github.com/marketplace/actions/pypi-publish#trusted-publishing. For some reason, I thought I added it to the PyPUG guide too but I don't see it there...

And the GitHub docs show the separation but don't call this out explicitly: https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-pypi. Though, William is trying to update this page with clarifications. The PR is in-flight but it takes many months to get GH to merge such changes so it'll take time, maybe another half a year...

youtux:
Ah I guess you mean this paragraph, right? Makes sense, will update.

> Pro tip: only set the id-token: write permission in the job that does publishing, not globally. Also, try to separate building from publishing — this makes sure that any scripts maliciously injected into the build or test environment won't be able to elevate privileges while flying under the radar.

PR #709, opened 2024-09-27
Test for slashes in Scenario Outline's Examples table

### LINKED ISSUE #655: Step definition is not found when using Scenario Outline and "/" characters in the Examples
Hello,

pytest-bdd: 7.0.0

I am having a simple Scenario Outline which is using table like this:

```
        | Popup    | Country  | Site                             | Locator                       |
        | False      | US           | https://my-site.com  | <selenium_locator>  |
```
In this case I am getting everytime: `pytest_bdd.exceptions.StepDefinitionNotFoundError` exception.
Removing `//` from the `https://my-site.com` seems to help and the step works perfectly fine in that case.

Are there any characters that are not allowed to be used in the Scenario Outline examples?

Thanks!
Marcin

jsa34:
Add test that forward slashes (in URLs for example) in scenario outline examples are handled properly by the parser.

Resolves #655


jsa34:
@youtux Not sure why this fails with py3.10, but my local 3.8 and the GH action for 3.12 was fine...

youtux:
> @youtux Not sure why this fails with py3.10, but my local 3.8 and the GH action for 3.12 was fine...

It's failing the `py3.10-pytest6.2-coverage` build, so when used in combination with pytest 6.2. We should just drop that old version of pytest

youtux:
Dropping compatibility with pytest 6.2 in https://github.com/pytest-dev/pytest-bdd/pull/711

jsa34:
Ah, thought I'd done that, but I had in a branch where this was also an issue but it hadn't been merged yet. My fault for raising too many MRs in a short space of time!

youtux (approved):

CHANGES.rst
```diff
@@ -11,7 +11,7 @@ Unreleased
 - Use `gherkin-official` parser to replace custom parsing logic. This will make pytest-bdd more compatible with the Gherkin specification.
 - Multiline steps must now always use triple-quotes for the additional lines.
 - All feature files must now use the keyword `Feature:` to be considered valid.
-- Tags can no longer have spaces (e.g. "@tag one" "@tag two" are no longer valid).
+- Tags can no longer have spaces (e.g., "@tag one" and "@tag two" are no longer valid).
```
youtux, line 14:
```suggestion
- Tags can no longer have spaces (e.g. "@tag one" and "@tag two" are no longer valid).
```

jsa34: Reverted :)

jsa34:
Looks like a recurring issue with xdist/test suite hanging until timeout clears it up.  Not sure how to fix this (unrelated to MR and happened in other branches)...

jsa34:
Looks like CI flakiness gods were in my favour this time.   Happy to merge, @youtux ?

youtux:
Yes!

PR #722, opened 2024-10-16
Do not strip comments from step names

youtux:
Otherwise we wouldn't be gherkin compliant

### DIFF
--- a/src/pytest_bdd/parser.py
+++ b/src/pytest_bdd/parser.py
@@ -20,21 +20,6 @@ from .gherkin_parser import get_gherkin_document
 from .types import STEP_TYPES
 
 STEP_PARAM_RE = re.compile(r"<(.+?)>")
-COMMENT_RE = re.compile(r"(^|(?<=\s))#")
-
-
-def strip_comments(line: str) -> str:
-    """Remove comments from a line of text.
-
-    Args:
-        line (str): The line of text from which to remove comments.
-
-    Returns:
-        str: The line of text without comments, with leading and trailing whitespace removed.
-    """
-    if res := COMMENT_RE.search(line):
-        line = line[: res.start()]
-    return line.strip()
 
 
 @dataclass(eq=False)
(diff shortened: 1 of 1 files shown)


youtux:
@jsa34 mind reviewing this?

jsa34:
Looks good and I agree - my first implementation was trying maintain compatibility, but makes total sense.

Just a couple of thoughts:

1.  Maybe a test to ensure # and trailing text don't get stripped?
2. There is a strict gherkin flag and tests - can we get rid of these now as we are always strict gherkin?
3. I haven't checked, but are there other places the old parser stripped comments in this way (such as example table entries)?  If so, perhaps the change log should reflect that not just comments in steps will no longer be stripped.

jsa34 (changes requested):
Looks good and I agree - my first implementation was trying maintain compatibility, but makes total sense.

Just a few thoughts:

1. Maybe a test to ensure # and trailing text don't get stripped?
1. There is a strict gherkin flag and tests - can we get rid of these now as we are always strict gherkin?
1. I haven't checked, but are there other places the old parser stripped comments in this way (such as example table entries)? If so, perhaps the change log should reflect that not just comments in steps will no longer be stripped.

youtux:
> 1. Maybe a test to ensure # and trailing text don't get stripped?
It's already tested in `test_scenario_comments`, but I will add testing for the step and and scenario names. Done in 48e51caf

> 2. There is a strict gherkin flag and tests - can we get rid of these now as we are always strict gherkin?
I'll check that and open another PR.

> 3. I haven't checked, but are there other places the old parser stripped comments in this way (such as example table entries)? If so, perhaps the change log should reflect that not just comments in steps will no longer be stripped.
It doesn't look like so.
Anyway I updated the CHANGES.rst to mention that now also Scenario and Feature names won't be stripped for comments (43d062b3)

jsa34 (approved):

Issue #397, opened 2020-11-11
Gherkin Localization Issue ?


arqeco:
Hello,

I've tried the following:

**login.feature**
```
#language: pt
#encoding: UTF-8
Funcionalidade: Login no Programa
    Cenário: O usuário ainda não é cadastrado
        Dado que o usuário esteja na tela de login
        Quando ele clicar no botão de Criar Conta
        Então ele deve ser levado para a tela de criação de conta
```
**test_login.py**
```
import pytest
 
from pytest_bdd import scenarios, given, when, then, parsers

scenarios("../features/login.feature")

@given("que o usuário esteja na tela de login")
def tela_login():
    assert True

@when("ele clicar no botão de Criar Conta")
def evento_criar_conta():
    assert True

@then("ele deve ser levado para a tela de criação de conta")
def tela_criacao_conta():
    assert True
```
This is the result:

```
ERROR step_defs/test_login.py - pytest_bdd.exceptions.NoScenariosFound
```

Is there support for Gherkin localization in pytest-bdd ?

https://cucumber.io/docs/gherkin/languages/

Thanks in advance,
Márcio

youtux:
Currently pytest-bdd is not localized, sorry for that.

arqeco:
Where in the pytest-bdd code someone must look for to try to implement localization ?

youtux:
The parsing logic is not completely straight forward. I started reimplementing the parser with the idea to make many things (including localization) easier, but I didn't manage to complete it.
I will probably start working on that again in the next month.

arqeco:
I began studying BDD a few days and the code above is my first try.

Gherkin natural language is great!

So I believe I'll be writing my feature files mixing the prefixes of each line in English and the rest of the line in Portuguese. 
Not pretty, but it will be simple to use the text editor to replace the prefixes with Portuguese equivalents when pytest-bdd is ready for localization.

Thank you for your work @youtux

arqeco:
Well, mixing languages in a feature file was not a good idea.

So now I am using this hook to create copies of the feature files, but with translated keywords, before the start of the tests.

**tests/conftest.py**
```python
import pytest
import pathlib

def pytest_sessionstart(session):
    keywords = {
        "Feature": ["Funcionalidade", "Característica", "Caracteristica"],
        "Background": ["Contexto", "Cenário de Fundo", "Cenario de Fundo", "Fundo"],
        "Scenario": ["Cenário", "Cenario"],
[...]

Issue #179, opened 2016-04-19
pytest-bdd should be internationalized


jangeroo:
Cucumber, Behave, and other BDD tools allow you to write your feature files in just about whatever language you want. It would be really handy to have the ability to use keywords other than the English versions.

A useful implementation would be one that allows anyone to create a new language file. Ideally it would also comes bundled with several language files for mainstream languages (i.e. French, Spanish, German, Russian, Chinese, etc.), though this is less useful than the ability to create a new custom language file.

olegpidsadnyi:
Do you mean having possibility to translate: "Feature, Background, Scenario, Given, When, Then, And"?

jangeroo:
Yes. That's exactly what I mean :)

Here's the wiki for this feature from the cucumber project https://github.com/cucumber/cucumber/wiki/Translations

gilmrjc:
+1

zaz600:
+1

AndersonFirmino:
+1

bubenkoff:
the whole thing can be done just by monkeypatching those
https://github.com/pytest-dev/pytest-bdd/blob/master/pytest_bdd/types.py
just make sure you monkeypatch it before calling scenario function

AndersonFirmino:
@bubenkoff Can you show me an example?

bubenkoff:
in the __init__py of the tests folder:

from pytest_bdd import types

types.FEATURE = "italian feature"
....

simpoir:
Actually, patching pytest_bdd.feature.STEP_PREFIXES would be slightly better, as you don't need to alter the types, but rather the parsing. And since a few of those translations are many-to-one-type.

e.g. according to https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json#L1138

```
from pytest_bdd.feature import STEP_PREFIXES
from pytest_bdd import types

STEP_PREFIXES[:] = []  # clear existing translations to avoid conflicts.
STEP_PREFIXES.append(("Quand ", types.WHEN))
STEP_PREFIXES.append(("Lorsque ", types.WHEN))
STEP_PREFIXES.append(("Lorsqu'", types.WHEN))
...
```

MichielPeeters7:
I have written this in the __init__.py file of the tests folder and I get a NoScenariosFound exception. Any idea what I did wrong?

TheoAndersen:
[...]

PR #725, opened 2024-10-19
Implement localisation support

youtux:
Fixes https://github.com/pytest-dev/pytest-bdd/issues/397
Supersedes https://github.com/pytest-dev/pytest-bdd/pull/345

Checklist:
- [x] Find all usages of `keyword` and check if we need to replace them with `keyword_type`

### DIFF
--- a/src/pytest_bdd/types.py
+++ b/src/pytest_bdd/types.py
@@ -12,3 +12,9 @@ WHEN: Literal["when"] = "when"
 THEN: Literal["then"] = "then"
 
 STEP_TYPES = (GIVEN, WHEN, THEN)
+
+STEP_TYPE_BY_PARSER_KEYWORD = {
+    "Context": GIVEN,
+    "Action": WHEN,
+    "Outcome": THEN,
+}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Unreleased
 - Dropped support for python 3.8. Supported python versions: 3.9, 3.10, 3.11, 3.12, 3.13.
 - Text after the `#` character is no longer stripped from the Scenario and Feature name.
 - Gherkin keyword aliases can now be used and correctly reported in json and terminal output (see `Keywords <https://cucumber.io/docs/gherkin/reference/#keywords>` for permitted list).
+- Added localization support. The language of the feature file can be specified using the `# language: <language>` directive at the beginning of the file.
 
 8.0.0b2
 ----------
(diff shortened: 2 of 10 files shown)


youtux:
@jsa34 mind reviewing this too?

jsa34:
There are some places where keywords are hardcoded, but I needed to resolve these for the rules PR, so this should fix this issue so that the gherkin terminal reporter and cucumber json out is using the desired keywords from the feature file, rather than assuming

jsa34:
We should include the language in the cucumber json output.

https://github.com/cucumber/compatibility-kit/blob/main/devkit/samples/minimal/minimal.feature
https://github.com/cucumber/compatibility-kit/blob/main/devkit/samples/minimal/minimal.feature.ndjson

This is example of a minimal out that is cucumber-spec-compliant, and as no language is specified, it defaults to "en" in this example, but obviously we should pass the value coming from the gherkin parser.

jsa34 (changes requested):
See comments in the MR

jsa34:
Also resolves #179

jsa34:
I also forgot an example and change log entry: it would be helpful to demonstrate how to specify the language in the feature file

src/pytest_bdd/types.py
jsa34, line 1: There is another type to map: Conjunction (And, But and *)

youtux: I know, but that's just a "tokenizer" type, not a semantic one, that's why I didn't include it here

youtux:
> We should include the language in the cucumber json output.
> 
> https://github.com/cucumber/compatibility-kit/blob/main/devkit/samples/minimal/minimal.feature https://github.com/cucumber/compatibility-kit/blob/main/devkit/samples/minimal/minimal.feature.ndjson
> 
> This is example of a minimal out that is cucumber-spec-compliant, and as no language is specified, it defaults to "en" in this example, but obviously we should pass the value coming from the gherkin parser.

Do you know if there is a specification somewhere for Cucumber JSON output?

youtux:
> I also forgot an example and change log entry: it would be helpful to demonstrate how to specify the language in the feature file

yes I forgot indeed, added in 342cf5edcfb516d4a4136ff6b7e722e7e48590b7

jsa34:
I think the parser test needs simplifying otherwise it will be a maintenance nightmare.
Perhaps just fix for this and then I will raise a PR to make the test more focussed.

youtux:
> I think the parser test needs simplifying otherwise it will be a maintenance nightmare. Perhaps just fix for this and then I will raise a PR to make the test more focussed.

Agree, but I think that once gherkin-official v30.0.0 is released, we can remove most of the classes in `gherkin_parser.py`, so let's wait for that one.

Can you check this again?

jsa34 (approved):

jsa34:
All good to me now :)

PR #343, opened 2020-01-09
Change in gherkin reporting to support multiple spoken languages

napolas:

PR #727, opened 2024-10-20
Don't hardcode keywords

jsa34:
Propogate keywords from gherkin parser to allow for internationalisation and using valid aliases for keywords

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Unreleased
 ----------
 - Dropped support for python 3.8. Supported python versions: 3.9, 3.10, 3.11, 3.12, 3.13.
 - Text after the `#` character is no longer stripped from the Scenario and Feature name.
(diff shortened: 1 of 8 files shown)


jsa34:
Along with #725 , this will supersede #343

youtux (commented):
Mind adding tests where we use an alternative to "Scenario Outline"? I think "Scenario Examples" should be an accepted alternative

jsa34:
> Mind adding tests where we use an alternative to "Scenario Outline"? I think "Scenario Examples" should be an accepted alternative

Done :)

youtux (dismissed):

youtux:
This will need a change log entry too right?

jsa34:
Wasn't quite sure how to word this, but added a change log entry.  Feel free to suggest clearer wording!

youtux (approved):

PR #728, opened 2024-10-20
Add rule feature

### LINKED ISSUE #511: Rule keyword
It'd be nice to have support for the [Rule](https://cucumber.io/docs/gherkin/reference/#rule) keyword and update for `Background`  to support rule level declarations.  Has anyone started on this?

For now global declarations of `Rule` at the top level are just treated as part of the feature description and passed through which is fine, but things break down when multiple rules are used to split a feature into it different scopes.

jsa34:
- Add changelog entry
- Add README entry
- Add tests
Resolves #511


jsa34:
@youtux  Created a new PR with the rule feature in as I messed up the rebasing!

README.rst
```diff
@@ -513,6 +513,43 @@ Example:
     def should_have_left_cucumbers(cucumbers, left):
         assert cucumbers["start"] - cucumbers["eat"] == left
 
+Rules
+-----
+
+In Gherkin, `Rules` allow you to group related scenarios or examples under a shared context.
+This is useful when you want to define different conditions or behaviours
+for multiple examples that follow a similar structure.
+You can use either `Scenario` or `Example` to define individual cases, as they are aliases and function identically.
```
youtux, line 522:
```suggestion
You can use either ``Scenario`` or ``Example`` to define individual cases, as they are aliases and function identically.
```

README.rst
```diff
@@ -513,6 +513,43 @@ Example:
...
+        Rule: A rule for invalid cases
+            Example: Invalid case
+                Given I have an invalid input
+                When I process the input
+                Then the result should be an error
+
+In Gherkin, `Scenario` and `Example` are interchangeable.
+You can use either keyword to define individual test cases.
+For example, the previous rule could have been written with the `Scenario` keyword instead of `Example`,
+and the behaviour would remain the same.
```
youtux, line 551: this is a repetition, it can be removed

src/pytest_bdd/generation.py
```diff
@@ -25,7 +26,7 @@
     from _pytest.main import Session
     from _pytest.python import Function
 
-    from .parser import Feature, ScenarioTemplate, Step
+    from .parser import ScenarioTemplate, Step
```
youtux, line 29: I think this can go in line 14

src/pytest_bdd/generation.py
```diff
@@ -100,18 +103,34 @@ def print_missing_code(scenarios: list[ScenarioTemplate], steps: list[Step]) ->
...
+            elif isinstance(feature_or_rule, Rule):
+                _feature: Feature = feature_or_rule.feature
+                tw.line(
+                    (
+                        f"Step {step} is not defined in the background of the rule "
+                        f'"{feature_or_rule.name}" in the file '
+                        f"{_feature.filename}:{step.line_number}"
+                    ),
+                    red=True,
+                )
```
youtux, line 133: is the only thing different the message? If so, let's have the message be conditionally defined from the `isinstance` check, rather than this huge blob.
something like:
```python
if isinstance(parent, Rule):
    parent_type = "rule"
    filename = parent
elif isinstance(parent, Feature):
    parent_type = "feature"
    filename = parent._filename
else:
    # raise NotImplementedError?
message =  (
    f"Step {step} is not defined in the background of the {parent_type} "
    f'"{parent.name}" in the file {filename}:{step.line_number}'
)
tw.line(message, ...)
```

src/pytest_bdd/gherkin_terminal_reporter.py
youtux, line 1: this looks like it should go in a separate PR

src/pytest_bdd/parser.py
```diff
@@ -134,14 +149,31 @@ def add_step(self, step: Step) -> None:
         step.scenario = self
         self._steps.append(step)
 
+    @property
+    def all_background_steps(self) -> list[Step]:
+        steps = []
+        # Add background steps from the feature
+        if self.feature.background:
+            steps.extend(self.feature.background.steps)
+        if self.rule and self.rule.background:
```
youtux, line 158:
```suggestion
        if self.rule is not None and self.rule.background is not None:
```

src/pytest_bdd/parser.py
```diff
@@ -165,7 +196,7 @@ def render(self, context: Mapping[str, Any]) -> Scenario:
             )
             for step in self._steps
```
youtux, line 197: I wonder, are steps in the background supposed to be rendered with the example rows as well? Gherkin docs don't mention it though

src/pytest_bdd/parser.py
```diff
@@ -293,12 +326,12 @@ class Background:
...
 
     Attributes:
-        feature (Feature): The feature to which this background belongs.
+        parent (Feature | Rule): The feature or rule to which this background belongs.
         line_number (int): The line number where the background starts in the file.
         steps (List[Step]): The list of steps in the background.
     """
 
-    feature: Feature
+    parent: Feature | Rule
```
youtux, line 334: maybe we should have 2 distinct fields here, `feature` and `rule`, since feature is always present, and rule is optional

src/pytest_bdd/parser.py
```diff
@@ -454,6 +491,22 @@ def parse(self) -> Feature:
...
+                    tags=self.get_tag_names(child.rule.tags),
+                    feature=feature,
+                )
+                if _backgrounds:
+                    _rule.add_background(self.parse_background(_backgrounds[0], _rule))
+                rule_scenarios = [c.scenario for c in child.rule.children if c.scenario]
+                if rule_scenarios:
+                    for rule_scenario in rule_scenarios:
+                        scenario = self.parse_scenario(rule_scenario, feature, _rule)
+                        feature.scenarios[scenario.name] = scenario
```
youtux, line 509: I wonder if there's a way to make this logic more straightforward.
There is a lot of mutations and it's hard to keep track of what's going on....

src/pytest_bdd/scenario.py
```diff
@@ -289,7 +289,8 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
                 example_parametrizations,
             )(scenario_wrapper)
 
-        for tag in templated_scenario.tags.union(feature.tags):
+        rule_tags = set() if templated_scenario.rule is None else templated_scenario.rule.tags
+        for tag in templated_scenario.tags.union(feature.tags).union(rule_tags):
```
youtux, line 293:
```suggestion
        for tag in templated_scenario.tags | feature.tags | rule_tags:
```

tests/feature/test_gherkin_terminal_reporter.py
```diff
@@ -297,3 +297,50 @@ def _():
...
+        """
+        ),
+    )
+    pytester.makepyfile(
+        test_gherkin=textwrap.dedent(
+            """\
+            from pytest_bdd import given, when, scenarios, then
+            @given("this is a step")
+            @when("this is a step")
+            @then("this is a step")
```
youtux, line 331:
```suggestion
            @step("this is a step")
```

tests/feature/test_rule_example_format.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_rule_example_format(pytester):
+    pytester.makefile(
+        ".feature",
+        rule_example=textwrap.dedent(
+            """\
+            Feature: Calculator
+
+              In order to perform basic arithmetic operations
```
youtux, line 11: let's keep using 4 spaces for indent

tests/feature/test_rule_example_format.py
```diff
@@ -0,0 +1,115 @@
...
+        def _(first_number, second_number):
+            return first_number + second_number
+
+        @when("I press subtract", target_fixture="result")
+        def _(first_number, second_number):
+            return first_number - second_number
+
+        @then(parsers.parse("the result should be {expected_result}"))
+        def _(result, expected_result):
+            assert result == int(expected_result)
```
youtux, line 104:
```suggestion
        @then(parsers.parse("the result should be {expected_result:d}"))
        def _(result, expected_result):
            assert result == expected_result
```

tests/feature/test_rule_example_format.py
```diff
@@ -0,0 +1,115 @@
...
+        def _():
+            pass
+
+        @given(parsers.parse("the first number is {first_number}"), target_fixture="first_number")
+        def _(first_number):
+            return int(first_number)
+
+        @given(parsers.parse("the second number is {second_number}"), target_fixture="second_number")
+        def _(second_number):
+            return int(second_number)
```
youtux, line 92:
```suggestion
        @given(parsers.parse("the second number is {second_number:d}"), target_fixture="second_number")
        def _(second_number):
            return second_number
```

tests/feature/test_rule_example_format.py
```diff
@@ -0,0 +1,115 @@
...
+        def _():
+            pass
+
+        @then("the screen turns on")
+        def _():
+            pass
+
+        @given(parsers.parse("the first number is {first_number}"), target_fixture="first_number")
+        def _(first_number):
+            return int(first_number)
```
youtux, line 88:
```suggestion
        @given(parsers.parse("the first number is {first_number:d}"), target_fixture="first_number")
        def _(first_number):
            return first_number
```

youtux (changes requested):

README.rst
jsa34: Oops!

src/pytest_bdd/parser.py
jsa34: Very good question.  I'm not 100% either about that.

src/pytest_bdd/parser.py
jsa34: I um-ed and ah-ed about this.  Will make the change :)

src/pytest_bdd/parser.py
youtux: let's leave it like that for now then

src/pytest_bdd/parser.py
jsa34: Extracted the logic into methods to make easier to read

src/pytest_bdd/parser.py
jsa34: Sorry, I misread this.

This needs to be only either Rule or Feature: A background can now be a rule-based background or a feature-based background.

See test case example:

>             Feature: Calculator
            
                In order to perform basic arithmetic operations
                As a user
                I want to use a calculator
            
                **Background:
                    Given I have got my calculator ready**
            
                Scenario: I check the calculator powers on
                    Given I press the power button
                    Then the screen turns on
            
                Rule: Addition
                    In order to add two numbers
                    As a user, I want the calculator to give me the sum.
            
                    **Background:
                        Given I check the add button is working**
            
                    Example: Adding two positive numbers
                        Given the first number is 3
                        And the second number is 5
                        When I press add
                        Then the result should be 8
            
                    Example: Adding a positive number and a negative number
                        Given the first number is 7
                        And the second number is -2
                        When I press add
                        Then the result should be 5


These two background elements need to identify what type they are.

That's the purpose of this field: is it a background group of steps under a feature or a rule?

jsa34:
Implemented all feedback.

Just this: https://github.com/pytest-dev/pytest-bdd/pull/728#discussion_r1815345859 I think needs to be as it is, so we know what type of background it is (under a feature or under rule), but may have misunderstood the suggestion!

Poogles:
Hero thanks for implementing this and the data tables work. <3

jsa34:
Pleasure! 

Glad it is hopefully helpful!

jsa34:
@youtux is there anything else after this that would be necessary for a V8 release? (I'm assuming https://github.com/pytest-dev/pytest-bdd/milestone/1 is our definition of what we want to achieve for v8?  If so, I can pick up the multiple examples table one and then i think that's it in terms of features)

youtux:
Unless there are outstanding issues I forgot about

youtux:
I'll try reviewing this in the weekend if I manage

jsa34:
No rush at all - Appreciated as always!

src/pytest_bdd/parser.py
```diff
@@ -135,14 +151,31 @@ def add_step(self, step: Step) -> None:
...
             List[Step]: A list of steps, including any background steps from the feature.
         """
-        return (self.feature.background.steps if self.feature.background else []) + self._steps
+        steps = []
+        # Add all background steps
+        steps.extend(self.all_background_steps)
+        # Add the scenario's own steps
+        steps.extend(self._steps)
+
+        return steps
```
youtux, line 178:
```suggestion
        return self.all_background_steps + self._steps
```

src/pytest_bdd/parser.py
youtux: if we are keeping this `parent` field just to show an error message, I'd say to drop it and instead display a generic error message like "Step ... is not defined".
Step generation functionality does not need to dictate so much of the models.

jsa34:
Committed responses to latest feedback :)

src/pytest_bdd/parser.py
```diff
@@ -100,6 +101,19 @@ def __bool__(self) -> bool:
         return bool(self.examples)
 
 
+@dataclass(eq=False)
+class Rule:
+    keyword: str
+    name: str
+    description: str
+    tags: set[str]
+    feature: Feature
```
youtux, line 110: Is this really needed?

jsa34: Which part?  The whole dataclass or the feature field?

youtux: The feature field

jsa34: Well spotted.  I probably used it when I first implemented somewhere, but the field is no longer used anywhere.  Removed.

jsa34:
Removed feature field from Rule

src/pytest_bdd/parser.py
```diff
@@ -453,9 +482,50 @@ def parse(self) -> Feature:
...
+            if child.background:
+                return self.parse_background(child.background)
+        return None
+
+    @staticmethod
+    def _extract_rule_scenarios(rule_data: GherkinRule) -> Generator[GherkinScenario]:
+        """Yield each scenario under a rule."""
+        for child in rule_data.children:
+            if child.scenario:
+                yield child.scenario
```
youtux, line 524: Shouldn't we `yield self._parse_and_add_scenario(child.scenario)` here instead?

Otherwise it breaks the convention where these `_extract_xxx(...)` methods return an object from the `parser.py` module (instead of a "raw" object from the gherkin_parser.py module)

jsa34: Good point!  Fixed :)

jsa34:
Feedback resolved :)

src/pytest_bdd/parser.py
```diff
@@ -453,9 +482,50 @@ def parse(self) -> Feature:
...
+            keyword=rule_data.keyword,
+            name=rule_data.name,
+            description=rule_data.description,
+            tags=self.get_tag_names(rule_data.tags),
+        )
+
+        # Add background if present within the rule
+        background = self._extract_rule_background(rule_data)
+        if background:
+            rule.add_background(background)
```
youtux, line 505: I guess we can avoid having this add_background method, and instead pass the background to the Rule constructor directly

src/pytest_bdd/reporting.py
```diff
@@ -126,6 +125,16 @@ def serialize(self) -> dict[str, Any]:
             },
         }
 
+        if scenario.rule:
+            serialized["rule"] = {
+                "keyword": scenario.rule.keyword,
+                "name": scenario.rule.name,
+                "description": scenario.rule.description,
+                "tags": sorted(scenario.rule.tags),
```
youtux, line 133: Why sorted?

jsa34:
Addressed feedback :)

youtux (dismissed):

src/pytest_bdd/gherkin_terminal_reporter.py
```diff
@@ -66,16 +67,27 @@ def pytest_runtest_logreport(self, report: TestReport) -> Any:
...
             self._tw.write(f"{report.scenario['feature']['keyword']}: ", **feature_markup)
             self._tw.write(report.scenario["feature"]["name"], **feature_markup)
             self._tw.write("\n")
-            self._tw.write(f"    {report.scenario['keyword']}: ", **scenario_markup)
+
+            if rule and rule["name"] != self.current_rule:
+                self._tw.write(f"  {rule['keyword']}: ", **rule_markup)
+                self._tw.write(rule["name"], **rule_markup)
+                self._tw.write("\n")
+                self.current_rule = rule["name"]
```
youtux, line 88: Can you make sure these are covered by tests?

jsa34: Added different verbosity levels to the existing test

youtux (approved):

jsa34:
@youtux tried rerunning the checks, but codecov consistently seems unhappy.  Not sure what's happening...

PR #731, opened 2024-10-27
Upload to PyPI the same dist that we tested on

youtux:


.github/workflows/main.yml
```diff
@@ -6,84 +6,116 @@ on:
...
-        tox run-parallel -f ${{ matrix.toxfactor }} --parallel-no-spinner --parallel-live
-        coverage combine
-        coverage xml
+      - name: Test with tox
+        continue-on-error: ${{ matrix.ignore-test-outcome }}
+        run: |
+          source .venv/bin/activate
+          coverage erase
+          # Using `installpkg dist/*.tar.gz` because we want to install the pre-built package (want to test against that)
+          tox run-parallel -f ${{ matrix.toxfactor }} --parallel-no-spinner --parallel-live --installpkg dist/*.tar.gz
```
webknjaz, line 108: Use `*.whl` instead. sdists are never installed directly. They'll always get built into wheels first.

webknjaz (commented):
It mostly seems fine except for the wheel thing. It's difficult to notice more things while reviewing from mobile with this many formatting changes. I recommend doing them separately in the interest of reviewability.

youtux:
> It mostly seems fine except for the wheel thing. It's difficult to notice more things while reviewing from mobile with this many formatting changes. I recommend doing them separately in the interest of reviewability.

This is the commit with the actual change: https://github.com/pytest-dev/pytest-bdd/commit/9797d536a4129923375223f6b3986e897bb40647

.github/workflows/main.yml
youtux: makes sense, done

webknjaz:
@youtux I recommend running `twine check --strict`. If you don't, `pypi-publish` will error out on metadata problems late in the process, which is an unpleasant experience.

youtux:
thanks, done in https://github.com/pytest-dev/pytest-bdd/pull/734

PR #738, opened 2024-11-11
Multiple example tables

### LINKED ISSUE #363: Scenario Outline to select examples based on user arguments?
Scenario outline has multiple example tables with env1,env2,2nv3 etc. Is there any way to collect only Examples:  env1 out of other Examples when user passes the argument env1. Basically the details are different among environments, so I cant use the markers as well. As markers on one example table might not work on another.

@Validate
Scenario Outline: Test to be executed in multiple env
Given the <username> and <password>

Examples: env1
|username|password|
|aa|nn|

Examples: env2
|username|password|
|aaq|nna|

Examples: env3
|username|password|
|aaa|n1n|

When I run pytest, I pass pytest --env=env1. Here I would like my pytest to execute only env1 examples.

I have following options
1) to place individual markers to each and run
2) to add new column and verify

Wondering if we have any good/reasonable approach.

Regards,
Tom.

### LINKED ISSUE #411: Tags support for Example sections
When I have Feature file like:
```gherkin
Feature: Foo
  Background:
    Given Step <bar>

    @cool_background_tag
    Examples:
    |bar|
    | 10|

    @nice_background_tag
    Examples:
    |bar|
    | 45|

  Scenario Outline: Fizz
    Given Step <buzz>

    @bad_scenario_tag
    Examples:
    |buzz|
    | 10|

    @ugly_scenario_tag
    Examples:
    |buzz|
    | 45|
```
and invoke *pytest* like:
```bash
# Runs Background only Examples for this tag:
pytest -m cool_background_tag
# Runs Scenario Outline only Examples for this tag:
pytest -m ugly_scenario_tag
# Other cases which also possible:
pytest -m "cool_background_tag and not bad_scenario_tag"
pytest -m "cool_background_tag or nice_background_tag"
```
*pytest-bdd* respects added tags

# NB:
Gherkin parser already has such model and it already used here: https://behave.readthedocs.io/en/latest/new_and_noteworthy_v1.2.6.html#index-0

jsa34:
Added multiple examples tables for scenario outlines.
Added tagging examples tables for filtering.

Supersedes #489 
Closes #411 
Closes #363


jsa34:
@youtux  would you be able to review and maybe we can get in the v8 release to close the milestone? :)

src/pytest_bdd/parser.py
```diff
@@ -413,7 +416,7 @@ def parse_scenario(self, scenario_data: GherkinScenario, feature: Feature) -> Sc
                     for row in example_data.table_body:
                         values = [cell.value or "" for cell in row.cells]
                         examples.add_example(values)
-                    scenario.examples = examples
+                scenario.examples.append(examples)
```
youtux, line 419: `ScenarioTemplate.examples` can be None, weird that mypy didn't catch this

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +304,29 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
-    else:
-        return None
+    parametrizations = []
+    has_multiple_examples = len(templated_scenario.examples) > 1
+
+    for example_id, examples in enumerate(templated_scenario.examples):
+        _tags: set = examples.tags or set()
+        example_marks = []
+        if _tags:
+            with warns(PytestUnknownMarkWarning, match=r"Unknown pytest\.mark\.\w+"):
```
youtux, line 314: `warns` asserts that the warning is raised. Why do we want that?

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +304,29 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
+                example_marks = [pytest.mark.__getattr__(tag) for tag in _tags]
```
youtux, line 315:
```suggestion
                example_marks = [getattr(pytest.mark, tag) for tag in _tags]
```

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +304,29 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
-        return None
-    if contexts := list(templated_scenario.examples.as_contexts()):
-        return [pytest.param(context, id="-".join(context.values())) for context in contexts]
-    else:
-        return None
+    parametrizations = []
+    has_multiple_examples = len(templated_scenario.examples) > 1
+
+    for example_id, examples in enumerate(templated_scenario.examples):
+        _tags: set = examples.tags or set()
```
youtux, line 311: no need to mark this variable as private with the leading underscore. can just call this `tags`

jsa34:
Sorry - I noticed it pulled in changes I didn't expect to terminal reporter.  Will revert :)

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +304,29 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
+        example_marks = []
+        if _tags:
+            with warns(PytestUnknownMarkWarning, match=r"Unknown pytest\.mark\.\w+"):
+                example_marks = [pytest.mark.__getattr__(tag) for tag in _tags]
+
+        for context in examples.as_contexts() or [{}]:
+            test_id = (
+                "-".join((str(example_id), *context.values())) if has_multiple_examples else "-".join(context.values())
```
youtux, line 319: why using the `example_id` to define the ID of the test?

tests/feature/test_outline.py
```diff
@@ -79,6 +79,56 @@ def test_outline(request):
...
+        textwrap.dedent(
+            """\
+        from pytest_bdd import scenarios
+
+        scenarios('outline_multi_example.feature')
+
+        """
+        )
+    )
+    result = pytester.runpytest("-s")
```
youtux, line 120: can we also run a test using `-k positive` to make sure that only those test items are executed?

src/pytest_bdd/scenario.py
jsa34: I started by building on what @olegpidsadnyi had committed in the superseded PR, and this was the approach he had.

tests/feature/test_tags.py
```diff
@@ -226,3 +226,47 @@ def _():
 
     result = pytester.runpytest("-m", "tag2", "-vv")
     result.assert_outcomes(passed=1, deselected=1)
+
+
+def test_tags_against_multiple_examples_tables(pytester):
```
youtux, line 231: to me this can be removed, and instead bring these assertions to the test you made in test_multiple_outlined.

Reason is that this test is really about the scenario outline functionality. I'm not even sure that this `test_tag.py` has a reason to exist. It's test items should just go in the test module where we assert the behaviour of Feature and Scenarios (and then Examples here)

src/pytest_bdd/parser.py
```diff
@@ -22,6 +22,18 @@
 STEP_PARAM_RE = re.compile(r"<(.+?)>")
 
 
+def get_tag_names(tag_data: list[GherkinTag]) -> set[str]:
```
youtux, line 25: not important: I'd prefer if this just accepted one single `GerkinTag`, and return one string. It should be up to the caller to call it for each  GherkinTag.

jsa34: This is maintaining both previous behaviour with old parser, and what the current logic is - this has always done this - I just moved it out of the scope of the feature file parser, as it needed to be available to the examples so that the pytest-bdd parser consistently represents all tags.

I think this should be a separate PR to change this behaviour?

src/pytest_bdd/scenario.py
jsa34: My bad - I forgot to replace my original workaround to stop warnings being emitted for unregistered marks!

src/pytest_bdd/scenario.py
youtux: Yes, I think we should not include the enumeration unless tests fail, can you try without example_id?

src/pytest_bdd/parser.py
youtux: Yes no worries, that's why I said optional

jsa34: Apologies - I misunderstood "not important" to mean the change I made is not important.

jsa34:
Changes pushed :)

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +305,26 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
-        return [pytest.param(context, id="-".join(context.values())) for context in contexts]
-    else:
-        return None
+    parametrizations = []
+
+    for examples in templated_scenario.examples:
+        tags: set = examples.tags or set()
+
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=PytestUnknownMarkWarning)
```
youtux, line 314: why do we need to silence warnings?

src/pytest_bdd/parser.py
youtux: this has still to be fixed

youtux: can we make `ScenarioTemplate.examples` not optional maybe? Or do we need to distinguish between no examples at all and empty examples table?

src/pytest_bdd/scenario.py
jsa34: My mistake - for some reason I thought the existing hook was suppressing warnings for marks registered from Gherkin tags.  I have now removed this code :)

src/pytest_bdd/parser.py
jsa34: Well, according to the official parser, a scenario is a scenario is a scenario.  All scenarios have an empty list of examples, but if the list is not empty, then it's a scenario outline.

youtux: I think there was a problem with copy paste in your last comment :)

youtux: So it looks like we can make it a non optional field, shall we do that ?

jsa34: The gherkin parser logic (which I have been conservative about modifying) used to use this None-ness I guess to distinguish between the two.  I am tempted to go the Gherkin view, which is why I originally modified the type to no longer be:

```
    examples: list[Examples] | None = field(default_factory=list[Examples])
```

but rather

```
    examples: list[Examples] = field(default_factory=list[Examples])
```

I added back the concept of optional based on your original comment.

youtux: Oh I didn't see you made it non nullable. That's why mypy didn't complain then. Yes please let's have it non nullable then

jsa34:
Resolved non-nullable examples list

src/pytest_bdd/scenario.py
```diff
@@ -303,12 +303,24 @@ def scenario_wrapper(request: FixtureRequest, _pytest_bdd_example: dict[str, str
...
-    else:
-        return None
+    parametrizations = []
+
+    for examples in templated_scenario.examples:
+        tags: set = examples.tags or set()
+
+        example_marks = [getattr(pytest.mark, tag) for tag in tags]
+
+        for context in examples.as_contexts() or [{}]:
```
youtux, line 313: Why `or [{}]`? Can't we coerce to an empty list instead?

jsa34: Again, copied from original impl and glossed over that this is no longer needed as a safeguard - as the examples can never be None

jsa34:
Feedback resolved :)

youtux (approved):

PR #744, opened 2024-11-29
Add test to assert that issue 389 is resolved.

### LINKED ISSUE #389: Unexpected behavior in scenario outline
Hi,

I have noticed unexpected behavior when using scenario outline. Here's a simple example to demonstrate it:

File `test.scenario`:
```
Feature: Test feature

  Scenario Outline: Scenario 1
    When Action "A" is taken
    And Some other action "<action>" is taken

    Examples: Actions
      | action |
      | B      |
```

File `steps.py`:
```
import pytest
from pytest_bdd import scenarios, given, when, then, parsers

scenarios('test.feature', example_converters={ 'action': str })

@when(parsers.parse('Action "{action}" is taken'))
def take_action1(action):
    print('take_action1:%s'%action)
    pass

@when('Some other action "<action>" is taken')
def take_action2(action):
    print('take_action2:%s'%action)
    pass
```

In this particular case I would expect to see the following in the output:
```
take_action1:A
take_action2:B
```

However, what I see is this:
```
take_action1:A
take_action2:A
```

Strangly this problem goes away if I rename `action` argument for one of the steps, for example:
```
@when(parsers.parse('Action "{action}" is taken'))
def take_action1(action):
    print('take_action1:%s'%action)
    pass

@when('Some other action "<action2>" is taken')
def take_action2(action2):
    print('take_action2:%s'%action2)
    pass
```
I have a feeling that arguments are somehow "cached" or memoized by name and was wondering is this expected behavior?

jsa34:
Tidy up imports in tests in the same file as new test written.

Resolves #389

### DIFF
--- a/tests/args/test_common.py
+++ b/tests/args/test_common.py
@@ -116,7 +116,7 @@ def test_string_steps_dont_take_precedence(pytester):
 
 
 def test_same_name_for_step_arg_and_example_parameter(pytester):
-    """Test that using the same name for step arg and example parameter as intended."""
+    """Test that using the same name for step arg and example parameter works as intended."""
     pytester.makefile(
         ".feature",
         step_args_examples_params=textwrap.dedent(
--- a/tests/args/test_common.py
+++ b/tests/args/test_common.py
@@ -74,7 +74,7 @@ def test_string_steps_dont_take_precedence(pytester):
     pytester.makeconftest(
         textwrap.dedent(
             """
-        from pytest_bdd import given, when, then, parsers
+        from pytest_bdd import given, when, then
         from pytest_bdd.utils import dump_obj
 
 
@@ -95,8 +95,7 @@ def test_string_steps_dont_take_precedence(pytester):
     pytester.makepyfile(
         textwrap.dedent(
             r"""
-        import pytest
-        from pytest_bdd import parsers, given, when, then, scenarios
+        from pytest_bdd import parsers, given, scenarios
         from pytest_bdd.utils import dump_obj
 
         scenarios("arguments.feature")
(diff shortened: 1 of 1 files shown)


tests/args/test_common.py
```diff
@@ -114,3 +113,49 @@ def _(value):
 
     [which] = collect_dumped_objects(result)
     assert which == "re"
+
+
+def test_same_name_for_step_arg_and_example_parameter(pytester):
+    """Test that using the same name for step arg and example parameter as intended."""
```
youtux, line 119: Missing the main clause in this sentence :)

youtux (approved):

tests/args/test_common.py
jsa34: Whoopsy!

tests/args/test_common.py
```diff
@@ -114,3 +113,49 @@ def _(value):
...
+        textwrap.dedent(
+            """
+        from pytest_bdd import scenarios, given, then, parsers
+
+        scenarios('.')
+
+
+        @given(parsers.parse('Action "{action}" is taken'))
+        def take_action1(action):
+            print(f'take_action1:{action}')
```
youtux, line 146: let's use `dumb_obj` and `collect_dumped_obj` instead, I made them for exactly this purpose

tests/args/test_common.py
```diff
@@ -114,3 +113,49 @@ def _(value):
 
     [which] = collect_dumped_objects(result)
     assert which == "re"
+
+
+def test_same_name_for_step_arg_and_example_parameter(pytester):
+    """Test that using the same name for step arg and example parameter works as intended."""
```
youtux, line 119: still not clear what we are testing here. Is it about params being filled from example tables vs from parsing the step name?
If so let's make it clearer in the docstring.

Maybe let's also rewrite the step names, like:
* Given foo gets value "A" by the step parser
* Given bar gets value <example_value> from the example table

This would be much clearer to me

youtux (changes requested):

tests/args/test_common.py
jsa34: I think the other test is actually testing the same thing and this can be closed

PR #745, opened 2024-11-29
Test that step args and example params with the same name do not redefine each other

### LINKED ISSUE #177: Conflict between Scenario Outline parameters and other steps' parameters
I'm not sure if this is  a feature or bug, but the behavior was a bit unexpected for me.  Consider the following scenario:

``` gherkin
Scenario Outline: Check for pytest-bdd variable re-use
    Given some <key> exists
    When I print <css_id>
    And I output "some value"
    And I echo <css_id>
    Then finish testing

    Examples:
    | key  | css_id |
    | foo  | ba     |
    | foo2 | bar2   |
```

Along with these steps:

``` python
@given("some <key> exists")
def some_key_exists(key):
    print("some {} exists".format(key))

@when('I print <css_id>')
def css_id(css_id):
    assert css_id in ('bar', 'bar2')

@when('I echo <css_id>')
def echo_val(css_id):
    assert css_id in ('bar', 'bar2')

@when(parsers.re('I output "(?P<css_id>.+)"'))
def i_output(css_id):
    assert css_id == 'some value'

@then('finish testing')
def i_output():
    print("finished")
    assert True
```

This results in:

``` bash
css_id = 'some value'
    @when('I echo <css_id>')
    def echo_val(css_id):
>       assert css_id in ('bar', 'bar2')
E       assert 'some value' in ('bar', 'bar2')

tests/functional/test_bugcheck.py:16: AssertionError
```

Since the step function for `And I output "some value"` uses `css_id` as its parameter name and that is the same name as one of the Scenario Outline examples tables, when `And I echo <css_id>` is run, it uses the `css_id` value from the previous step (`And I output "some value"`).  

Is this the desired behavior?  I understand that re-using parameters across steps is desired (in case an object change needs to pass between steps), but when scenario outlines are involved, maybe this shouldn't apply? If this behavior is desirable, what's a use case for it? And do you have any favorite "best practices" to help avoid this potential conflict?

jsa34:
Resolves #177

### DIFF
--- a/tests/feature/test_outline.py
+++ b/tests/feature/test_outline.py
@@ -334,9 +334,9 @@ def test_variable_reuse(pytester):
             """\
             Feature: Example parameters reuse
                 Scenario Outline: Check for example parameter re-use
-                    Given I get <param>
-                    When I set param to "other"
-                    Then I check <param>
+                    Given the param is initially set from the example table as <param>
+                    When a step arg of the same name is set to "other"
+                    Then the param is still set from the example table as <param>
 
                     Examples:
                         | param |
@@ -355,17 +355,17 @@ def test_variable_reuse(pytester):
             scenarios('outline.feature')
 
 
-            @given(parsers.parse('I get {param}'))
+            @given(parsers.parse('the param is initially set from the example table as {param}'))
             def _(param):
                 dump_obj(("param1", param))
 
 
-            @when(parsers.re('I set param to "(?P<param>.+)"'))
+            @when(parsers.re('a step arg of the same name is set to "(?P<param>.+)"'))
             def _(param):
                 dump_obj(("param2", param))
 
 
-            @then(parsers.parse('I check {param}'))
+            @then(parsers.parse('the param is still set from the example table as {param}'))
             def _(param):
                 dump_obj(("param3", param))
 
--- a/tests/feature/test_outline.py
+++ b/tests/feature/test_outline.py
@@ -323,7 +323,10 @@ def test_forward_slash_in_params(pytester):
 
 
 def test_variable_reuse(pytester):
-    """Test example parameter reuse."""
+    """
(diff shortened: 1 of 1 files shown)


youtux (changes requested):
sorry, forgot to submit the review

tests/feature/test_outline.py
```diff
@@ -320,3 +320,70 @@ def _(site):
...
+                    Given some <key> exists
+                    When I print <css_id>
+                    And I output "some value"
+                    And I echo <css_id>
+                    Then finish testing
+
+                    Examples:
+                        | key  | css_id |
+                        | foo  | bar    |
+                        | foo2 | bar2   |
```
youtux, line 346: can we compress this test to the minimal amount of steps necessary?
Also, let's use more of `dump_obj` and `collect_dumped_objects` instead of asserting in `Then` steps.
This way if the `Then` functionality is broken and passes even though it shouldn't, we would still be doing the assertions.

jsa34: Good point - my bad for just copying the test from the issue as a starting point!

tests/feature/test_outline.py
```diff
@@ -320,3 +320,58 @@ def _(site):
...
+
+
+            @given(parsers.parse('I get {param}'))
+            def _(param):
+                dump_obj(("param1", param))
+
+
+            @when(parsers.re('I set param to "(?P<param>.+)"'))
+            def _(param):
+                dump_obj(("param2", param))
```
youtux, line 365: maybe better use steps like:
* `'param is initially set from the example table to "{value}"'`
* `'param is then by the step to to "{value}"'`

I'd find it easier to grasp

youtux (approved):
some optional comments if you feel like fixing them, otherwise feel free to merge

youtux (approved):

PR #746, opened 2024-11-30
Fix call to pytest.mark.usefixtures without arguments

daara-s:
This will raise a warning in an [upcoming change in pytest](https://github.com/pytest-dev/pytest/pull/13008).

### DIFF
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -297,7 +297,7 @@ def _get_scenario_decorator(
             config.hook.pytest_bdd_apply_tag(tag=tag, function=scenario_wrapper)
 
         scenario_wrapper.__doc__ = f"{feature_name}: {scenario_name}"
-        scenario_wrapper.__scenario__ = templated_scenario
+        scenario_wrapper.__scenario__ = templated_scenario  # type: ignore[attr-defined]
         return cast(Callable[P, T], scenario_wrapper)
 
     return decorator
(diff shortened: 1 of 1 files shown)


jsa34:
I think we need to either define a sort of "ScenarioFunction" protocol for mypy to be happy and expect __scenario__, or some short term ignoring of the mypy error, unless I am missing something, @youtux ?

The-Compiler:
It's not quite clear to me how this change leads to mypy suddenly seeing the type of `scenario_wrapper`. It was `Any` before, and now mypy correctly sees that it's actually `def (request: _pytest.fixtures.FixtureRequest, _pytest_bdd_example: builtins.dict[builtins.str, builtins.str]) -> Any` (aka `Callable[[FixtureRequest, dict[str, str]], Any]`).

Given the decorator *is* typed, this should already have been the case before I believe?

As this isn't really related to this PR and we do a `return cast(Callable[P, T], scenario_wrapper)` at the end already anyways, in the scope of this PR it seems reasonable to me to ignore the error.

youtux (dismissed):

youtux (approved):

youtux:
> I think we need to either introduce a sort of "ScenarioFunction" protocol for mypy to be happy and expect **scenario**, or some short term ignoring of the mypy error, unless I am missing something, @youtux ?

Actually I wanted to fix that in a different way. I started the work in https://github.com/pytest-dev/pytest-bdd/pull/658, but never got to finish it...

jsa34:
> > I think we need to either introduce a sort of "ScenarioFunction" protocol for mypy to be happy and expect **scenario**, or some short term ignoring of the mypy error, unless I am missing something, @youtux ?
> 
> Actually I wanted to fix that in a different way. I started the work in #658, but never got to finish it...

Ah, yes!  I remember this now!

PR #747, opened 2024-11-30
Empty example table entries with converter to None documentation

### LINKED ISSUE #551: New pytest bdd 6.0.1 doesnt take null values from the Examples within the scenario outline
Steps to reproduce

1. Within a Pytest feature file try to have a scenario outline
2. Within example (just take example for a login page) as discussed
3.Scenario Outline: Login functionality with Negative password  and valid email
        Given User opens login in page
        When the User enters <invalid_password> into password and valid credential on the other textbox
        Then verify wrong_password will display a <warning_message> under the textbox
        Examples:
        |  invalid_password     |    warning_message                       |
        |                                    |    Password cannot be empty        |
        
        Step definition for the above scenario :
        def test_login_invalidPassword_Cases():
        pass

@when(parsers.parse('the User enters {invalid_password} into {textbox} and valid credential on the other textbox'))
def login_with_negative_password(textbox, invalid_password):
    Login_functions.fn_InvalidLogin(textbox, invalid_password)
        
  4. In the run time of the test ,the null value is getting as " " (double qoutes)

jsa34:
Add documentation for parser converter for empty strings in example tables.

Resolves #551

### DIFF
--- a/tests/feature/test_outline_empty_values.py
+++ b/tests/feature/test_outline_empty_values.py
@@ -4,6 +4,28 @@ import textwrap
 
 from pytest_bdd.utils import collect_dumped_objects
 
+STEPS = """\
+from pytest_bdd import given, when, then, parsers
+from pytest_bdd.utils import dump_obj
+
+# Using `parsers.re` so that we can match empty values
+
+@given(parsers.re("there are (?P<start>.*?) cucumbers"))
+def _(start):
+    dump_obj(start)
+
+
+@when(parsers.re("I eat (?P<eat>.*?) cucumbers"))
+def _(eat):
+    dump_obj(eat)
+
+
+@then(parsers.re("I should have (?P<left>.*?) cucumbers"))
+def _(left):
+    dump_obj(left)
+
+"""
+
 
 def test_scenario_with_empty_example_values(pytester):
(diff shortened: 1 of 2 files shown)


tests/feature/test_outline_empty_values.py
```diff
@@ -62,3 +62,65 @@ def test_outline():
     result = pytester.runpytest("-s")
     result.assert_outcomes(passed=1)
     assert collect_dumped_objects(result) == ["#", "", ""]
+
+
+def test_scenario_with_empty_example_values_none_transformer(pytester):
```
youtux, line 67: please add a docstring explaining what's the issue we are covering here

jsa34: Good point - I'll also add an example in the README as I realised this is probably the helpful place for people wanting to know how to do it

jsa34:
@youtux Could you review and see if this is clear/correct/helpful?

youtux (approved):
do we really need a test for this? isn't this tested somewhere else already? Maybe not this very example, but the fact that converters... basically work.

jsa34:
Good point.  I'll just add to the docs

youtux (approved):

PR #748, opened 2024-11-30
Raise an error if a step defines reserved argument names (`datatable`, `docstring`)

youtux:
Let's prevent headaches in the future and let's directly raise an error in case a step includes an argument like `datatable` or `docstring` that are also used to inject DataTable and Docstring objects.

### DIFF
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -28,7 +28,7 @@ from . import exceptions
 from .compat import getfixturedefs, inject_fixture
 from .feature import get_feature, get_features
 from .steps import StepFunctionContext, get_step_fixture_name
-from .utils import CONFIG_STACK, get_args, get_caller_module_locals, get_caller_module_path
+from .utils import CONFIG_STACK, get_args, get_caller_module_locals, get_caller_module_path, identity
 
 if TYPE_CHECKING:
     from _pytest.mark.structures import ParameterSet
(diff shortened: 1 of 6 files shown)


youtux:
I made this change while reviewing that PR. I think we should merge this first, as it makes the logic for getting parsed args more self contained.

src/pytest_bdd/scenario.py
```diff
@@ -172,6 +175,27 @@ def get_step_function(request: FixtureRequest, step: Step) -> StepFunctionContex
             return None
 
 
+def parse_step_arguments(step: Step, context: StepFunctionContext) -> dict[str, object]:
+    """Parse step arguments."""
+    parsed_args = context.parser.parse_arguments(step.name)
+
+    assert parsed_args is not None, (
+        f"Unexpected `NoneType` returned from " f"parse_arguments(...) in parser: {context.parser!r}"
+    )
```
jsa34, line 184: I've always found this a bit odd (maybe just me!)

Can we refactor this to match the rest of the exception raising styles, like:

```
    parsed_args = context.parser.parse_arguments(step.name)
    if parsed_args is None:
        raise ValueError(
            f"`parse_arguments` returned `None` for step {step.name!r}. "
            f"Parser: {context.parser!r}"
        )
```

Or raise the new custom StepImplementationError?

jsa34:
Sorry - pressed the wrong button - never knew that icon on mobile requested a review!

src/pytest_bdd/scenario.py
youtux: Makes sense, I'll take that in a separate PR if you don't mind though

jsa34 (approved):

PR #749, opened 2024-12-01
Add scenario description to JSON and add test for JSON correctly populating the description from a feature

jsa34:
Adds scenario description to json output.

The test for the json output didn't check the feature description field was populated.  I have added a value here so we can be sure this is being set.

Resolves #351

PR #751, opened 2024-12-05
Fix `null` returned in Cucumber JSON `scenario.description`

youtux:
Fix regression of https://github.com/pytest-dev/pytest-bdd/pull/749 where we would serialize the `scenario.description` as `null` instead of the empty string.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,7 @@ Removed
 
 Fixed
 +++++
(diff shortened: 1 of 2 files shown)


youtux:
@jsa34 FYI

youtux:
It turns out there was no bug, only our `Scenario` and `ScenarioTemplate` typing is incorrect. Will fix them in https://github.com/pytest-dev/pytest-bdd/pull/658

PR #758, opened 2024-12-31
Migrate `black`, `flake8`, `isort`, `pyupgrade` linters and formatters to `ruff`

kieran-ryan:
### 🤔 What's changed?

- Migrated linters [`flake8`](https://flake8.pycqa.org/), [`isort`](https://pycqa.github.io/isort/), [`pydocstyle`](https://pycodestyle.pycqa.org/en/latest/) and [`pyupgrade`](https://github.com/asottile/pyupgrade) to [`ruff`](https://docs.astral.sh/ruff)
- Migrated formatter [`black`](https://black.readthedocs.io/en/stable/) to `ruff`
- Froze pre-commit dependency versions to commit hashes
- Applied and enforced `from __future__ import annotations` on all Python modules
- Updated docstring type hints to match `__future__.annotations` syntax e.g. `List` to `list`, `Optional` to `... | None`
- Applied [`codespell`](https://pypi.org/project/codespell/1.9.1/) spelling corrections

### ⚡️ What's your motivation? 

- Migrating to ruff provides the following enhancements:
  - ⚡️Extremely fast - 42x faster than flake8 in ruff's benchmark - and even faster with built-in caching
    - See pytest-bdd pre-commit run [_before_ (4.6s)](https://results.pre-commit.ci/run/github/9094192/1733662754.3Alsch-XRqe7VDtJavRIJA) and [_after_ (2.6s)](https://results.pre-commit.ci/run/github/9094192/1735768054.FREynhL-TBSpUE7ADGxgJA)
  - 📏 [Hundreds of additional linting rules](https://docs.astral.sh/ruff/rules)
  - 🐁 Reduces linting and formatting dependencies (`black`, `flake8`, `flake8-bugbear`, `flake8-pyproject`, `mccabe`, `pycodestyle`, `pyflakes`, `isort`, `pyupgrade`, `tokenize-rt`) down to `ruff` - including "Drop-in parity with flake8, isort, and black"
  - 🔧 Fix support, for automatic error correction
  - ⌨️ First-party [editor integrations](https://docs.astral.sh/ruff/integrations/) for [VS Code](https://github.com/astral-sh/ruff-vscode) and [more](https://docs.astral.sh/ruff/editors/setup/)
  - 🛠️ Native `pyproject.toml` support - without requiring dependencies such as `flake8-pyproject` - and can be read by IDE extensions and is monorepo-friendly, with [hierarchical and cascading configuration](https://docs.astral.sh/ruff/configuration/#config-file-discovery). Centralises configuration rather than splitting across pre-commit, pyproject.toml, IDE settings, etc.
  - 🤝 Simplifies contributing experience and running linting and formatting locally or through developer tools
- [pytest has adopted ruff](https://github.com/pytest-dev/pytest/pull/11901) 
- Appears to be at least [some level of ruff adoption amongst pytest-bdd contributors](https://github.com/pytest-dev/pytest-bdd/pull/653#discussion_r1394971473)
- Freezing pre-commit hashes is better for security (matches cucumber/gherkin#326)

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

### ♻️ Anything particular you want feedback on?

- Whether suitable to include a recommended VSCode workspace extension or should omit

### References

- [flake8 error codes](https://flake8.pycqa.org/en/latest/user/error-codes.html)
- [pycodestyle error codes](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) (installed with flake8)
- [ruff rules](https://docs.astral.sh/ruff/rules)

### DIFF
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -384,7 +384,7 @@ def scenario(
         scenario = feature.scenarios[scenario_name]
     except KeyError:
         feature_name = feature.name or "[Empty]"
-        raise exceptions.ScenarioNotFound(
+        raise exceptions.ScenarioNotFound(  # noqa: B904
             f'Scenario "{scenario_name}" in feature "{feature_name}" in {feature.filename} is not found.'
         )
 
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -188,9 +188,9 @@ def parse_step_arguments(step: Step, context: StepFunctionContext) -> dict[str, 
     """Parse step arguments."""
     parsed_args = context.parser.parse_arguments(step.name)
 
-    assert parsed_args is not None, (
-        f"Unexpected `NoneType` returned from " f"parse_arguments(...) in parser: {context.parser!r}"
-    )
+    assert (
+        parsed_args is not None
+    ), f"Unexpected `NoneType` returned from parse_arguments(...) in parser: {context.parser!r}"
 
     reserved_args = set(parsed_args.keys()) & STEP_ARGUMENTS_RESERVED_NAMES
     if reserved_args:
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -384,9 +384,9 @@ def scenario(
         scenario = feature.scenarios[scenario_name]
     except KeyError:
         feature_name = feature.name or "[Empty]"
-        raise exceptions.ScenarioNotFound(  # noqa: B904
+        raise exceptions.ScenarioNotFound(
             f'Scenario "{scenario_name}" in feature "{feature_name}" in {feature.filename} is not found.'
-        )
+        ) from None
 
     return _get_scenario_decorator(
         feature=feature, feature_name=feature_name, templated_scenario=scenario, scenario_name=scenario_name
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,33 +1,16 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
--   repo: https://github.com/psf/black
-    # If you update the version here, also update it in tox.ini (py*-pytestlatest-linters)
-    rev: 24.10.0
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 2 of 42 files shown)


Pierre-Sassoulas (commented):
Looks great !

jsa34:
@youtux we talked bout ruff before but I seem to remember you were keen to stick to psf-supported solutions, such as black.

I'll defer to @youtux

src/pytest_bdd/scenario.py
```diff
@@ -384,7 +385,7 @@ def scenario(
         scenario = feature.scenarios[scenario_name]
     except KeyError:
         feature_name = feature.name or "[Empty]"
-        raise exceptions.ScenarioNotFound(
+        raise exceptions.ScenarioNotFound(  # noqa: B904
```
youtux, line 388: we should do `raise ... from None` instead of suppressing the error from ruff

youtux (commented):
I'm willing to give it a shot to use `ruff`, given also [`pytest` started using it](https://github.com/pytest-dev/pytest/pull/11901).

There's just one thing I'd like to see addressed: I'd want to do as few import as possible inside the `if TYPE_CHECKING:` block, ideally none at all. The imports we do within that block are pretty much modules that should be already loaded in the interpreter, so the overhead of importing them is negligible.

src/pytest_bdd/scenario.py
kieran-ryan: Cheers! Applied

kieran-ryan:
> I'm willing to give it a shot to use `ruff`, given also [`pytest` started using it](https://github.com/pytest-dev/pytest/pull/11901).

Awesome @youtux!

> There's just one thing I'd like to see addressed: I'd want to do as few import as possible inside the `if TYPE_CHECKING:` block, ideally none at all. The imports we do within that block are pretty much modules that should be already loaded in the interpreter, so the overhead of importing them is negligible.

Happy to change; clarifying my assumptions on scope: revert imports migrated into type checking blocks and revert adding the [flake8-type-checking-tc (TC)](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) linting rules (or a subset)?

youtux:
Yes, both, thank you

kieran-ryan:
> Yes, both, thank you

Dropped associated commit and updated pull request description - some of these imports may be dropped with #759 through associated gherkin parser enhancements in any case, so good shout 👍

youtux (approved):

PR #759, opened 2025-01-01
Remove Gherkin local dataclasses

jsa34:
Remove the custom definition of the gherkin document as dataclasses in pytest-bdd.

Bump the minimum supported version of gherkin-official to v30 to support this and remove the warnings in python 3.13+


src/pytest_bdd/gherkin_parser.py
```diff
@@ -56,264 +49,56 @@
...
-        return cls(
-            feature=Feature.from_dict(data["feature"]),
-            comments=[Comment.from_dict(comment) for comment in data["comments"]],
-        )
-
-
-def _to_raw_string(normal_string: str) -> str:
-    return normal_string.replace("\\", "\\\\")
+def replace_datatable_values(document: GherkinDocument) -> None:
+    """Replace all cell values in DataTables within a GherkinDocument using _to_raw_string."""
```
youtux, line 53: why is this necessary? Maybe it should be done by `gherkin-official`?

src/pytest_bdd/parser.py
```diff
@@ -521,18 +522,18 @@ def _parse_and_add_rule(self, rule_data: GherkinRule, feature: Feature) -> None:
...
     def _extract_rule_scenarios(
         self, rule_data: GherkinRule, feature: Feature, rule: Rule
     ) -> Generator[ScenarioTemplate]:
         """Yield each parsed scenario under a rule."""
-        for child in rule_data.children:
-            if child.scenario:
-                yield self.parse_scenario(child.scenario, feature, rule)
+        for child in rule_data["children"]:
+            if "scenario" in child:
+                yield self.parse_scenario(child["scenario"], feature, rule)  # type: ignore[typeddict-item]
```
youtux, line 536: we should try avoiding `# type: ignore`. What's the issue here? Maybe we need to cast `child` to a specific type?

src/pytest_bdd/gherkin_parser.py
youtux: also, I'd rather avoid trying to patch datatables here at all, and instead handle it in `src/pytest_bdd/scenario.py`, since it's one place there, but many places here

tests/parser/test_errors.py
youtux, line 1: shouldn't we keep testing the error messages?

youtux (commented):
maybe we should change `gherkin-official` to return classes instead of dicts now. WDYT?

jsa34:
> maybe we should change `gherkin-official` to return classes instead of dicts now. WDYT?

Good point - ideally, I guess it should return the gherkin messages format (which will be dataclasses) so should wait to do this until that is resolved.  I'll mark this as draft until then but you have rightly pointed out workarounds that should probably also be backported to gherkin-official.

PR #789, opened 2025-12-06
[packaging] Set the licence files metadata for new versions of setuptools/poetry

Pierre-Sassoulas:
Trying to fix the pipeline on main so we can merge the depandabot upgrade.

PR #790, opened 2025-12-06
Update Poetry to version 2.2.1 in CI workflow

youtux:
* Updated the CI workflow and build system to use Poetry version 2.2.1. 
* Update twine to the latest version
* Add latest gherkin-offical versions and pythons to the test matrix.

This fixes an issue with `twine` failing due to unknown license metadata.


Pierre-Sassoulas:
It seems there's the same error than in #789 ? (both in the pipeline and once merged in main)

Pierre-Sassoulas:
(Well in #789 I skipped the failing test so the pipeline is green but it was failing before that)

youtux:
CI in the master branch is failing because of unexpected behaviour from the latest gherkin_official parser versions. It will need to be addressed separately.

PR #795, opened 2026-02-24
fix(wallet): necessary changes to programmatic step generation example

abreumateus:
Fix on missing variables, additional imports, and relative path.


abreumateus:
I agree that the Copilot suggestions align with the README structure rather than the standard pytest-bdd conventions. I appreciate the feedback.

PR #796, opened 2026-03-03
feat: use scenarios() shortcut in code generation template

### LINKED ISSUE #535: Scenario generation should use `scenarios` shortcut
The current template generates a test item for each scenario. We should just use the `scenarios` shortcut instead.

toroleapinc:
## Summary

Update the code generation template to use the `scenarios()` shortcut instead of individual `@scenario()` decorators.

Closes #535

## Changes

Modified `src/pytest_bdd/templates/test.py.mak` to:
- Import `scenarios` instead of `scenario`
- Generate `scenarios('feature_file.feature')` calls instead of individual `@scenario(...)` decorated test functions
- Automatically deduplicate feature file paths

## Before

```python
from pytest_bdd import given, scenario, then, when

@scenario('features/login.feature', 'Successful login')
def test_successful_login(): ...

@scenario('features/login.feature', 'Failed login')
def test_failed_login(): ...
```

## After

```python
from pytest_bdd import given, scenarios, then, when

scenarios('features/login.feature')
```


src/pytest_bdd/templates/test.py.mak
```diff
@@ -1,22 +1,22 @@
+<%
+    feature_paths = sorted(set(s.feature.rel_filename for s in scenarios))
```
youtux: - Shouldn't we use the `features` car instead of `scenarios`?

toroleapinc: Good catch, updated to use `features` directly.

src/pytest_bdd/templates/test.py.mak
```diff
@@ -1,22 +1,22 @@
+<%
+    feature_paths = sorted(set(f.rel_filename for f in features))
+%>\
```
youtux, line 3: this should go right before line 14

youtux (approved):

toroleapinc:
Pushed ed581b5 — updated the expected output strings in `tests/scripts/test_generate.py` (`test_generate`, `test_generate_with_quotes`, `test_unicode_characters`) to match the new `scenarios()` codegen output. Verified locally on Python 3.13: the 3 codegen tests pass, and the only remaining test failure I see (`test_step_outside_scenario_or_background_error`) reproduces on `upstream/master` too, so it's pre-existing and not from this PR.

Looks like the CI run is currently blocked on `action_required` (workflow approval for the fork). Could a maintainer approve it when you get a chance? Thanks.

Pierre-Sassoulas (approved):
Previously approved by youtux and all the nex changes come from main.

youtux:
@Pierre-Sassoulas why did you merge this before my comment was addressed?

Also, with this change do we preserve idempotency, in case the autogen script is invoked twice?

Pierre-Sassoulas:
Sorry that's my bad, do you want me to revert ?

youtux:
Yes please, let's then address my comments and then we can ship it

Pierre-Sassoulas:
I added a test in the new MR regarding idempotency.

PR #811, opened 2026-06-08
Fix pytest.fail() not captured in cucumber JSON step output

jackhenderson101:
pytest.fail() raises _pytest.outcomes.Failed which inherits from BaseException, not Exception. The except Exception handler in _execute_step_function was bypassing the pytest_bdd_step_error hook, leaving the step's failed flag as False and causing all steps to appear as "passed" in the JSON output even when the test failed.

### DIFF
--- a/src/pytest_bdd/scenario.py
+++ b/src/pytest_bdd/scenario.py
@@ -24,6 +24,7 @@ from weakref import WeakKeyDictionary
 
 import pytest
 from _pytest.fixtures import FixtureDef, FixtureManager, FixtureRequest, call_fixture_func
+from _pytest.outcomes import Failed
 
 from . import exceptions
 from .compat import getfixturedefs, inject_fixture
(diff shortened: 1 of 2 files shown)


hiSandog:
Capturing `pytest.fail()` in cucumber JSON is the right direction, but the edge case is hook ordering around step teardown. It would be good to include a scenario where `pytest.fail()` happens inside the step body and another where failure is raised from a fixture/teardown path, then assert the JSON has the failed status and useful message in both cases. That keeps the report format reliable for CI consumers.

src/pytest_bdd/scenario.py
```diff
@@ -247,7 +247,7 @@ def _execute_step_function(
         # so that we can allow "yield" statements in it
         return_value = call_fixture_func(fixturefunc=context.step_func, request=request, kwargs=kwargs)
 
-    except Exception as exception:
+    except BaseException as exception:
```
youtux: How about instead adding the catch for Failed instead?

```suggestion
    except (Exception, Failed) as exception:
```

youtux (changes requested):

src/pytest_bdd/scenario.py
jackhenderson101: Have narrowed the exception clause to only Exception and pytest.Failed

youtux (approved):

PR #819, opened 2026-07-02
Upgrade depencencies to fix cves, add support for python 3.14 and 3.15

Pierre-Sassoulas:
Fix dependabot security alerts by upgrading locked dev dependencies with ``poetry update urllib3 cryptography starlette filelock requests pytest``. Python 3.9 keeps older fallback pins (latest releases dropped 3.9 support), but all locked versions are free of known advisories (checked against OSV). Dropping python 3.9 would probably be a good idea at this point.

Also add Python 3.14 and 3.15-dev to the CI matrix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,15 @@ jobs:
             toxfactor: py3.13
             ignore-typecheck-outcome: false
             ignore-test-outcome: false
+          - python-version: "3.14"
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas:
@youtux I upgraded the pipeline to fix the CVEs warned by dependabot

youtux (approved):

PR #820, opened 2026-07-02
Fix over-quoting of backslashes in table cells

### LINKED ISSUE #769: Too much quoting of backslashes in datatables
**Describe the bug**

Backslashes in datatables in a feature file get converted to double backslashes in the datatable fixture dictionary.

**To Reproduce**

Given a feature file `test.feature`:

```
Feature: Quoting

    Scenario: Backslahes are quoted correctly
        Then expect backslash in datatable
          | \ |
```

and a test file:

```
from pytest_bdd import scenario, then

@scenario('test.feature', 'Backslahes are quoted correctly')
def test_backslash():
    pass


@then("expect backslash in datatable")
def _(datatable):
    assert datatable[0][0] == '\\'
```

The test fails with the following output:

```
=========== short test summary info =================
FAILED test_backslash.py::test_backslash - AssertionError: assert '\\\\' == '\\'
  
  - \
  + \\
=========== 1 failed in 0.05s ===========================
```

**Expected behavior**

The content of the datatable should appear exactly as in the feature file: a single backslash becomes a single backslash.

**Version**
 - pytest version: 8.3.5
 - pytest-bdd version: 8.1.0
 - OS: Linux (Debian bookworm)

golikovichev:
Fixes #769

## Problem

Backslashes in datatable and Examples table cells were doubled. A cell written as a single backslash reached the step as two:

```gherkin
Then expect backslash in datatable
  | \ |
```

```python
@then("expect backslash in datatable")
def _(datatable):
    assert datatable[0][0] == "\\"   # failed: value was "\\\\"
```

## Cause

`gherkin_parser.Cell.from_dict` ran every cell value through a helper:

```python
def _to_raw_string(normal_string: str) -> str:
    return normal_string.replace("\\", "\\\\")
```

The `gherkin` library already resolves cell escaping per the Gherkin spec before pytest-bdd sees the value (`\\` -> `\`, `\|` -> `|`, `\n` -> newline, a lone `\` stays `\`). The extra pass doubled every backslash a second time. This looks like the leftover compatibility workaround mentioned in the issue thread.

## Fix

Deliver the cell value exactly as `gherkin` resolves it and drop the now-unused helper (it had a single call site).

## Backwards compatibility

This is a behaviour change, not a pure internal fix. Cells now carry the Gherkin-resolved value with no extra doubling. Anyone who relied on the old doubled output (for example asserting `\\` for a single authored backslash, or writing `C:\\\\Users` to survive the old double pass) will need to drop that workaround. Values now match what the feature author wrote and what other Gherkin implementations produce.

## Tests

- New `tests/datatable/test_datatable.py::test_datatable_preserves_backslashes` covering a lone backslash, an escaped `\\`, an escaped pipe `\|`, and a Windows path `C:\Users\John`. Expected values were checked against the raw `gherkin` parser output.
- Updated the two `test_outline_with_escaped_pipes` expectations that encoded the old doubled output (`bork      \\` now decodes to a single trailing backslash; `bork    \\\|` now decodes to `bork    \|`).
- `CHANGES.rst`: entry under Unreleased / Fixed.

Full suite passes locally except for pre-existing Windows-only path-separator failures in `test_generate` / `test_migrate` / `test_feature_base_dir` that also fail on an unmodified checkout and are unrelated to this change.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,7 @@ Removed
 
 Fixed
 +++++
+* Backslashes in datatable and examples table cells are no longer over-quoted. A cell containing a single backslash now reaches the step as a single backslash, matching the Gherkin escaping rules. `#769 <https://github.com/pytest-dev/pytest-bdd/issues/769>`_
 * Made type annotations stronger and removed most of the ``typing.Any`` usages and ``# type: ignore`` annotations. `#658 <https://github.com/pytest-dev/pytest-bdd/pull/658>`_
 
 Security
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ Removed
 
 Fixed
 +++++
-* Backslashes in datatable and examples table cells are no longer over-quoted. A cell containing a single backslash now reaches the step as a single backslash, matching the Gherkin escaping rules. `#769 <https://github.com/pytest-dev/pytest-bdd/issues/769>`_
+* Backslashes in datatable and examples table cells are no longer over-quoted. A cell containing a single backslash now reaches the step as a single backslash, matching the Gherkin escaping rules. If you compensated by doubling backslashes in feature files, undo that. `#769 <https://github.com/pytest-dev/pytest-bdd/issues/769>`_
 * Made type annotations stronger and removed most of the ``typing.Any`` usages and ``# type: ignore`` annotations. `#658 <https://github.com/pytest-dev/pytest-bdd/pull/658>`_
 * Empty docstrings are now correctly forwarded to step functions as an empty string instead of being silently dropped, which previously caused pytest to report a missing ``docstring`` fixture. `#809 <https://github.com/pytest-dev/pytest-bdd/issues/809>`_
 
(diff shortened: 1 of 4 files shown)


CHANGES.rst
```diff
@@ -30,6 +30,7 @@ Removed
 
 Fixed
 +++++
+* Backslashes in datatable and examples table cells are no longer over-quoted. A cell containing a single backslash now reaches the step as a single backslash, matching the Gherkin escaping rules. `#769 <https://github.com/pytest-dev/pytest-bdd/issues/769>`_
```
Pierre-Sassoulas:
```suggestion
* Backslashes in datatable and examples table cells are no longer over-quoted. A cell containing a single backslash now reaches the step as a single backslash, matching the Gherkin escaping rules. if you compensated by doubling backslashes in feature files, undo that `#769 <https://github.com/pytest-dev/pytest-bdd/issues/769>`_
```

Pierre-Sassoulas (commented):
LGTM, @jsa34 do you have an opinion ?

CHANGES.rst
golikovichev: Good call, applied. Thanks for the review.

jsa34:
@Pierre-Sassoulas I remember reluctantly (trying to keep the change as minimal for regression as possible) adding this in for backwards compatibility, but fully support removing this custom escaping and undesirable behaviour.

Pierre-Sassoulas (approved):

PR #821, opened 2026-07-03
Drop support for Python 3.9 (EOL since October 2025)

Pierre-Sassoulas:
The remaining 15 dependabot alerts could not be resolved while poetry.lock still solved for 3.9: every patched release (urllib3 2.7.0, cryptography 49.0.0, starlette 1.3.1, pytest 9.0.3, filelock 3.20.3, requests 2.33.0) requires Python >= 3.10, so the lock kept vulnerable 3.9 fallback pins. Follow-up to #819.

Also bumps ruff to `target-version = py310` and applies the new lint fixes.


poetry.lock
```diff
@@ -926,27 +723,13 @@ perf = ["ipython"]
...
-description = "brain-dead simple config-ini parsing"
-optional = false
-python-versions = ">=3.8"
-groups = ["main", "dev"]
-markers = "python_version == \"3.9\""
-files = [
-    {file = "iniconfig-2.1.0-py3-none-any.whl", hash = "sha256:9deba5723312380e77435581c6bf4935c94cbfab9b1ed33ef8d238ea168eb760"},
-    {file = "iniconfig-2.1.0.tar.gz", hash = "sha256:3abbd2e30b36733fee78f9c7f7308f2d0050e88f0087fd25c2645f63c773e1c7"},
-]
-
```
Pierre-Sassoulas, line 941: There was two entries for some value (as you can see there's an iniconfig entry for 2.3.0 below), this is why there's a lot of removal.

Pierre-Sassoulas:
@youtux the only way to fix the CVE was to drop python 3.9, or it's impossible to upgrade (urllib 2.6.3 does not support python 3.9 for example)

pyproject.toml
```diff
@@ -74,7 +73,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.ruff]
 line-length = 120
-target-version = "py39"
+target-version = "py310"
```
webknjaz: This is inferred from the project table above by default. No need to duplicate the same info.
```suggestion
```

Pierre-Sassoulas: Indeed, thank you for the review !

Pierre-Sassoulas:
Would also need a setting change to not require the python 3.9 job anymore

webknjaz:
> Would also need a setting change to not require the python 3.9 job anymore

Integrate alls-green instead?

Pierre-Sassoulas:
I added 3.14 and removed 3.9 in required job, I don't know about all green, could be a separate MR.

poetry.lock
```diff
@@ -394,135 +378,13 @@ files = [
 ]
 markers = {main = "sys_platform == \"win32\"", build = "os_name == \"nt\""}
 
-[[package]]
-name = "coverage"
```
youtux, line 398: Why?

poetry.lock
```diff
@@ -1666,39 +1381,13 @@ files = [
     {file = "pyproject_hooks-1.2.0.tar.gz", hash = "sha256:1e859bd5c40fae9448642dd871adf459e5e2084186e8d2c2a79a824c970da1f8"},
 ]
 
-[[package]]
-name = "pytest"
```
youtux, line 1670: Why?

poetry.lock
```diff
@@ -2183,33 +1811,7 @@ files = [
     {file = "tomli-2.3.0-py3-none-any.whl", hash = "sha256:e95b1af3c5b07d9e643909b5abbec77cd9f1217e6d0bca72b0234736b9fb1f1b"},
     {file = "tomli-2.3.0.tar.gz", hash = "sha256:64be704a875d2a59753d80ee8a533c3fe183e3f06807ff7dc2232938ccb01549"},
 ]
-markers = {main = "python_version < \"3.11\"", build = "python_version < \"3.11\"", dev = "python_full_version <= \"3.11.0a6\""}
-
-[[package]]
-name = "tox"
```
youtux, line 2189: Why?

pyproject.toml
```diff
@@ -74,7 +73,6 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.ruff]
 line-length = 120
-target-version = "py39"
```
youtux, line 77: We should replace with 3.10 target

youtux (changes requested):

poetry.lock
Pierre-Sassoulas: https://github.com/pytest-dev/pytest-bdd/pull/821#discussion_r3517771745

 (same for all the removal)

pyproject.toml
Pierre-Sassoulas: I initially did, but see the previous review ruff is using a sensible default value.

poetry.lock
Pierre-Sassoulas: Due to python 3.9 requiring specific version for each interpreter.

youtux: But we dropped Python 3.9...

poetry.lock
```diff
@@ -394,135 +378,13 @@ files = [
...
-toml = ["tomli ; python_full_version <= \"3.11.0a6\""]
-
 [[package]]
 name = "coverage"
 version = "7.12.0"
 description = "Code coverage measurement for Python"
 optional = false
 python-versions = ">=3.10"
 groups = ["dev"]
-markers = "python_version >= \"3.10\""
```
Pierre-Sassoulas, line 525: As you can see here we're dropping the python version specifier for coverage 7.12.0, and we removed the coverage specific to python 3.9 (7.10.7), so we only have coverage 7.12.0 for all supported interpreters now.

poetry.lock
Pierre-Sassoulas: Yes, so we're dropping the coverage version that was used only for python 3.9, see here for the remaining coverage version: https://github.com/pytest-dev/pytest-bdd/pull/821#discussion_r3525340464

youtux (approved):

PR #822, opened 2026-07-03
Use a match statement to dispatch on gherkin Child fields

Pierre-Sassoulas:
Opinionated change following #821. A gherkin Child has exactly one of background/rule/scenario set; class patterns make that dispatch explicit and replace the truthiness checks with type matching.

### DIFF
--- a/src/pytest_bdd/parser.py
+++ b/src/pytest_bdd/parser.py
@@ -4,11 +4,10 @@ import copy
 import os.path
 import re
 import textwrap
+import warnings
 from collections import OrderedDict
 from collections.abc import Generator, Iterable, Mapping, Sequence
 from dataclasses import dataclass, field
-
-from typing_extensions import assert_never
 
 from .exceptions import StepError
 from .gherkin_parser import Background as GherkinBackground
@@ -501,12 +500,14 @@ class FeatureParser:
                     self._parse_and_add_rule(rule, feature)
                 case GherkinScenario() as scenario:
                     self._parse_and_add_scenario(scenario, feature)
-                case None:  # pragma: no cover
-                    # An empty Child (a child kind newer than this gherkin version);
-                    # skip it, like unknown children were skipped before the match.
-                    pass
-                case unreachable:  # pragma: no cover
-                    assert_never(unreachable)
+                case _:
+                    # An empty Child, or a child kind newer than this gherkin version;
+                    # skip it and warn instead of crashing, to keep working against
+                    # future gherkin releases that add new child types.
+                    warnings.warn(
+                        f"Unknown gherkin child skipped in {self.rel_filename!r}; consider upgrading pytest-bdd.",
+                        stacklevel=2,
+                    )
 
         return feature
 
--- a/src/pytest_bdd/parser.py
+++ b/src/pytest_bdd/parser.py
@@ -8,9 +8,11 @@ from collections import OrderedDict
 from collections.abc import Generator, Iterable, Mapping, Sequence
 from dataclasses import dataclass, field
 
+from typing_extensions import assert_never
+
 from .exceptions import StepError
 from .gherkin_parser import Background as GherkinBackground
-from .gherkin_parser import Child, DataTable, GherkinDocument, get_gherkin_document
+from .gherkin_parser import DataTable, GherkinDocument, get_gherkin_document
 from .gherkin_parser import Feature as GherkinFeature
 from .gherkin_parser import Rule as GherkinRule
 from .gherkin_parser import Scenario as GherkinScenario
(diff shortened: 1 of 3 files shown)


src/pytest_bdd/parser.py
```diff
@@ -492,12 +492,13 @@ def parse(self) -> Feature:
...
 
         for child in feature_data.children:
-            if child.background:
-                feature.background = self.parse_background(child.background)
-            elif child.rule:
-                self._parse_and_add_rule(child.rule, feature)
-            elif child.scenario:
-                self._parse_and_add_scenario(child.scenario, feature)
+            match child:
+                case Child(background=GherkinBackground() as background):
```
youtux: Is it needed the `as foo` part?

tests/parser/test_parser.py
```diff
@@ -851,3 +853,35 @@ def test_parser():
...
+    )
+
+    real_get_gherkin_document = parser_module.get_gherkin_document
+
+    def get_gherkin_document_with_unknown_child(abs_filename: str, encoding: str = "utf-8") -> GherkinDocument:
+        document = real_get_gherkin_document(abs_filename, encoding)
+        document.feature.children.append(Child())
+        return document
+
+    monkeypatch.setattr(parser_module, "get_gherkin_document", get_gherkin_document_with_unknown_child)
```
youtux: Do we really need to mock? Not clear to me what's going on and why we need this, and how it relates to the change made

youtux (changes requested):

tests/parser/test_parser.py
Pierre-Sassoulas: It's done in order to cover all the code branches to make codecov happy. With current gherkin code never produce a Child with none of background/rule/scenario set, so no real .feature file can exercise the fall-through branch. (I can drop it if you want)

src/pytest_bdd/parser.py
Pierre-Sassoulas: The class pattern GherkinBackground() only tests the type, it doesn't bind. We could use child.background though.

Pierre-Sassoulas: But it creates complication for type checkers usually.

src/pytest_bdd/parser.py
```diff
@@ -492,12 +492,13 @@ def parse(self) -> Feature:
         )
 
         for child in feature_data.children:
-            if child.background:
-                feature.background = self.parse_background(child.background)
-            elif child.rule:
-                self._parse_and_add_rule(child.rule, feature)
-            elif child.scenario:
-                self._parse_and_add_scenario(child.scenario, feature)
+            match child:
```
youtux: Can we have a `case _ as unreachable: assert_never(unreachable)`?

https://typing.python.org/en/latest/guides/unreachable.html

youtux (changes requested):

youtux:
If we adopt the assert_never suggestion, we shouldn't need the test, as mypy would cover it

Pierre-Sassoulas:
mypy doesn't understand it (yet?) sadly. ``child.background or child.rule or child.scenario`` gives mypy ``Background | Rule | Scenario | None`` so I had to handle the None too.

src/pytest_bdd/parser.py
```diff
@@ -492,12 +494,19 @@ def parse(self) -> Feature:
...
+                    feature.background = self.parse_background(background)
+                case GherkinRule() as rule:
+                    self._parse_and_add_rule(rule, feature)
+                case GherkinScenario() as scenario:
+                    self._parse_and_add_scenario(scenario, feature)
+                case None:  # pragma: no cover
+                    # An empty Child (a child kind newer than this gherkin version);
+                    # skip it, like unknown children were skipped before the match.
+                    pass
+                case unreachable:  # pragma: no cover
```
youtux: Although I wonder, if they add new types, maybe we should gracefully ignore them, but add a warning message? But our CI wouldn’t break so we wouldn’t really see it… but i think maybe it’s for the best, so that we have less maintenance burden.

Pierre-Sassoulas: Yeah, I think the warning message make sense. If it triggers it's going to be a lot of duplicate issue opened, but it's probably going to be duplicate issue in any case, just less obvious. Let me know if I should add it

youtux: Yeah I'd say to remove the assert_never and instead have a catch-all case that emits a warning

tests/parser/test_parser.py
youtux: ~Can we test that we emit a warning  now?~ Also, let's use unittest.patch.object instead, in context manager mode (with-statement)

Pierre-Sassoulas: I did that in the latest commit, the coverage is still not green tough, I can ivnestigate if you want.

src/pytest_bdd/parser.py
```diff
@@ -492,12 +493,21 @@ def parse(self) -> Feature:
...
+                    self._parse_and_add_rule(rule, feature)
+                case GherkinScenario() as scenario:
+                    self._parse_and_add_scenario(scenario, feature)
+                case _:
+                    # An empty Child, or a child kind newer than this gherkin version;
+                    # skip it and warn instead of crashing, to keep working against
+                    # future gherkin releases that add new child types.
+                    warnings.warn(
+                        f"Unknown gherkin child skipped in {self.rel_filename!r}; consider upgrading pytest-bdd.",
+                        stacklevel=2,
```
youtux, line 509: Why?

tests/parser/test_parser.py
```diff
@@ -851,3 +856,36 @@ def test_parser():
...
+
+    real_get_gherkin_document = parser_module.get_gherkin_document
+
+    def get_gherkin_document_with_unknown_child(abs_filename: str, encoding: str = "utf-8") -> GherkinDocument:
+        document = real_get_gherkin_document(abs_filename, encoding)
+        document.feature.children.append(Child())
+        return document
+
+    with patch.object(parser_module, "get_gherkin_document", get_gherkin_document_with_unknown_child):
+        with pytest.warns(UserWarning, match="Unknown gherkin child"):
```
youtux, line 887: No need for a nested with, we can have 2 context managers in the same with-statement

PR #825, opened 2026-07-05
feat: use scenarios() shortcut in code generation template (take 2)

### LINKED ISSUE #535: Scenario generation should use `scenarios` shortcut

Pierre-Sassoulas:
Re-applies #796 (reverted in #824 because it was merged before review comments were addressed), with @youtux's review comments addressed:

- **`features` vs `scenarios` var**: kept the fix from the original PR — the template iterates over `features` directly.
- **Template block placement**: the `<% feature_paths = ... %>` computation now sits right before the loop that uses it, inside the `% if features:` block, instead of at the top of the file ([comment](https://github.com/pytest-dev/pytest-bdd/pull/796#discussion_r2927044049)).
- **Idempotency** ([comment](https://github.com/pytest-dev/pytest-bdd/pull/796#issuecomment-4886573785)): yes, it is preserved, and now covered by a test (`test_generate_is_idempotent`):
  - invoking the autogen script twice yields byte-identical output (`feature_paths` is a `sorted(set(...))`, so deterministic);
  - applying the generated code twice to the same module does not duplicate tests, because `scenarios()` skips scenarios already bound in the caller module.

Also adds the changelog entry that was missing from the original PR.

Original work by @toroleapinc.

Closes #535


src/pytest_bdd/templates/test.py.mak
```diff
@@ -3,20 +3,20 @@
...
 )
 
-
-% endif
-% for scenario in sorted(scenarios, key=lambda scenario: scenario.name):
-@scenario('${scenario.feature.rel_filename}', ${ make_string_literal(scenario.name)})
-def test_${ make_python_name(scenario.name)}():
-    ${make_python_docstring(scenario.name)}
+<%
+    feature_paths = sorted(set(f.rel_filename for f in features))
```
youtux, line 12: This can be inlined

tests/scripts/test_generate.py
```diff
@@ -157,6 +151,37 @@ def _():
...
+                    Given I have a step
+
+                Scenario: Second scenario
+                    Given I have a step
+            """
+        ),
+    )
+
+    first = pytester.run("pytest-bdd", "generate", "idempotency.feature")
+    second = pytester.run("pytest-bdd", "generate", "idempotency.feature")
```
youtux, line 176: Oh I thought this script was writing to a file... if not then this test is irrelevant

PR #827, opened 2026-07-26
fix: avoid deprecated nodeid argument to _register_fixture

### LINKED ISSUE #823: Deprecation warnings with pytest 9.1.1 and python 3.14.0
Image

We ran into deprecation warnings

`Passing nodeid to _register_fixture is deprecated. Pass node instead for fixture scoping.`

`pytest_bdd/compat.py:28 | PytestRemovedIn10Warning`

`_pytest/fixtures.py:1986 | PytestRemovedIn10Warning`

Sanjays2402:
Closes #823

`inject_fixture` calls `FixtureManager._register_fixture(nodeid=...)`, which pytest 9.1 deprecated in favour of `node=`, so every scenario using a `target_fixture` emits `PytestRemovedIn10Warning` (and fails outright in suites that error on deprecation warnings). Now passes `node=request.node` on pytest >= 9.1 and keeps `nodeid` on older versions, matching the existing version-branching in `compat.py`.

Regression test added in `tests/steps/test_given.py`; it fails on master and passes with the fix. Full suite is unchanged on pytest 9.1, and `tests/steps` + `tests/feature` still pass on pytest 9.0.


youtux (dismissed):

Sanjays2402:
Closed the patch-coverage gap Codecov flagged on compat.py: added tests/test_compat.py with direct unit tests for both version branches of inject_fixture — node= on pytest 9.1+ and the legacy nodeid= path (via a mocked pytest_version, since no CI job runs pytest < 9.1). Both pass locally along with the existing given-injection regression test.

youtux:
can you add pytest 9.1 to the `tox.ini` matrix and delete the tests you added? it should fail without your fix right?

PR #807, opened 2026-04-24
Show scenario tags in gherkin terminal reporter

baobaodawang-creater:
Adds scenario tag output to the gherkin terminal reporter when tags are present.

This prints tags in the requested format, for example:

(tags: R1.1, R1.2)

The change only affects tagged scenarios. Untagged scenarios keep the existing output.

Tests added:
- tagged scenario output for -v
- tagged scenario output for -vv
- guard that untagged scenario output does not include a tags line

Validation:
- python -m pytest tests/feature/test_gherkin_terminal_reporter.py -q --maxfail=1
- python -m pytest tests/feature -q --maxfail=1

Closes #476

PR #830, opened 2026-09-06
Display scenario tags in gherkin terminal reporter

### LINKED ISSUE #476: print tags in gherkin-terminal-reporter

ryux1:
Closes #476.

Scenario tags are already serialized into the report context, but the Gherkin terminal reporter never renders them. This adds a small shared writer and emits `(tags: ...)` after the scenario result in both verbose modes. Untagged scenario output remains unchanged.

Validation:
- `pytest -q`: 144 passed, 1 skipped
- focused reporter suite: 16 passed
- Ruff check and format on the changed Python files
- MyPy: clean across 18 source files

The repository-wide Ruff run still reports the pre-existing `ISC004` finding in `tests/generation/test_generate_missing.py:80`; this PR does not touch that file.


ryux1:
Closing as a duplicate. I missed the existing #807 and #477 because my initial search was limited to references to issue #476. Sorry for the noise; the existing contributors should retain ownership of this work.

